PR #1, opened 2024-01-19
Update README.md

atwardowski:
Fix PORT env variable in docker compose example


thijsvanloef:
Small typo, good catch!

PR #6, opened 2024-01-19
Add ADMIN_PASSWORD environment variable

Twinki14:
# Motivations
Being able to configure the admin password would be quite helpful

# Modifications
- Add `.gitattributes`, to ensure the scripts use LF line endings
- Add `ADMIN_PASSWORD` to the list of environment variables, which adds `--adminpassword` to the server arguments, this has been verified to work locally

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,6 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     && apt-get clean \
     && rm -rf /var/lib/apt/lists/*
 
-
 ENV PORT=8211 \
     PLAYERS=16 \
     MULTITHREADING=false \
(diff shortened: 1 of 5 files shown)


thijsvanloef:
Thank you for your PR, will take a look 👀

Dockerfile
```diff
@@ -15,6 +14,7 @@ ENV PORT=8211 \
     PUBLIC_PORT= \
     SERVER_PASSWORD= \
     SERVER_NAME= \
+    ADMIN_PASSWORLD= \
```
thijsvanloef, line 17: Small typo, should be ADMIN_PASSWORD

thijsvanloef (changes requested):

thijsvanloef (approved):

PR #18, opened 2024-01-21
Add linux/arm64 build arg to docker/build-push-action

### LINKED ISSUE #10: May I have ARM 64 version?
I tried your docker image but not work. It shows that exec format error. I guess my Linux is ARM64. So, May I get the ARM version for this docker image? Thank you very much.

alex4108:
This might fix #10

### DIFF
--- a/.github/workflows/docker-hub.yml
+++ b/.github/workflows/docker-hub.yml
@@ -32,5 +32,6 @@ jobs:
         uses: docker/build-push-action@v2
         with:
           push: true
+          platforms: linux/amd64,linux/arm64
           tags: |
             ${{ steps.meta.outputs.tags }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -35,4 +35,5 @@ jobs:
         uses: docker/build-push-action@v2
         with:
           push: true
+          platforms: linux/amd64,linux/arm64
           tags: ${{ steps.meta.outputs.tags }}


thijsvanloef:
Before changing the release workflow, I will be testing a container build and server start on an ARM64 machine to make sure that it works

thijsvanloef:
> Hi! The steamcmd docker image this image is based on, does not support ARM64. Running this container on ARM devices (Raspberry PI, Oracle Ampere A1) would require an overhaul of the Dockerfile.

_Originally posted by @thijsvanloef in https://github.com/thijsvanloef/palworld-server-docker/issues/10#issuecomment-1902583902_

thijsvanloef:
@alex4108 Do you intend to work more on this PR? This is currently not a working solution

alex4108:
Heya!  If this doesn't work I don't have many other ideas - steam and palworld themselves need to ship arm64 support before we can try anything else imo.

alex4108:
I checked https://www.pocketpair.jp/aboutus and found the CEO might have a github account here: https://github.com/history-of

I sent an email to that account and to info@pocketpair.jp to see if they want to schedule arm64 support for future work.  Pointed them at this PR

tanpengsccd:
Here is ARM64 for SteamCMD : https://github.com/TeriyakiGod/steamcmd-docker-arm64 
And somte make it dose work: https://hub.docker.com/r/czy0612/palworld-server  , I test it , I work for my Oracle ARM 4C16G100G.
but can't apply ENV for that Image.

Issue #40, opened 2024-01-21
Add ENV Variable to Enable RCON and set RCON port


xHyperElectric:
It would be awesome if there was an option to enable RCON and set the RCON port.

**Describe the solution you'd like**
An ENV Variable option to enable RCON and set its port. i.e. `ENABLE_RCON = true` and `RCON_PORT = 25575`. RCON can be set in "Pal/Saved/Config/LinuxServer/PalWorldSettings.ini" with these values from the .ini: "RCONEnabled=False, RCONPort=25575".

xHyperElectric:
I would love to be able to edit any value in the "Pal/Saved/Config/LinuxServer/PalWorldSettings.ini" file, but I'm not sure the best way to do it. Maybe with some guidance, I could submit a PR. Would the best way to handle it just be `cd /.../Pal/Saved/Config/LinuxServer/` and something like: `sed -i 's/RCONEnabled=False/RCONEnabled=true/' PalWorldSettings.ini`?

thijsvanloef:
@xHyperElectric Probably just `sed` the RCONEnabled in the `DefaultPalServerSettings.ini` and after all the changes have been made, copy it to the `PalServerSettings.ini` in `/.../Pal/Saved/Config/LinuxServer/`

xHyperElectric:
What would be the best way to allow all of the settings in PalWorldSettings.ini to be changed in the docker file?

thijsvanloef:
@xHyperElectric Feel free to open a PR even if you are unsure, we can always take a look at it together :)

thijsvanloef:
@xHyperElectric most user friendly way would be via environment variables

xHyperElectric:
So is it practical to make an environment variable option for all 62 options?

thijsvanloef:
@xHyperElectric no not really practical, that is partly the reason i've yet to work on it, because there isn't really any other "user friendly" way

xHyperElectric:
I think the best way to do it is to add a second Volume for Server Config Storage and have the PalWorldSettings.ini file in there and it overwrites the one in `/.../Pal/Saved/Config/LinuxServer/`. I think it's [recommended to use a bind-mount](https://stackoverflow.com/questions/61202608/recommended-way-to-provide-configuration-files-to-docker-containers) for this but I am by no means a docker expert.

Twinki14:
[...]

PR #42, opened 2024-01-21
Add ENABLE_RCON ENV option and RCON_PORT ENV option

xHyperElectric:
I believe I changed all instances of "PORT" to "GAME_PORT" to delineate from "RCON_PORT" and Query Port, I hope I didn't miss any.
This is my first experience writing docker files or using sed, so please check it to make sure I didn't make any mistakes.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,5 +30,5 @@ WORKDIR /home/steam/server
 HEALTHCHECK --start-period=5m \
     CMD pgrep "PalServer-Linux" > /dev/null || exit 1
 
-EXPOSE ${PORT}
+EXPOSE ${GAME_PORT} ${RCON_PORT}
 ENTRYPOINT ["/home/steam/server/init.sh"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     && apt-get clean \
     && rm -rf /var/lib/apt/lists/*
 
-ENV PORT= \
+ENV GAME_PORT= \
     PUID=1000 \
     PGID=1000 \
     PLAYERS= \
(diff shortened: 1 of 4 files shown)


xHyperElectric:
Double check start.sh as I have 0 experience with sed and got help from ChatGPT. I'm afraid the ENABLE_RCON option has no safety for if RCONEnabled is already set to true in the file. Idk what sed does when it fails

xHyperElectric:
I believe the file has to at minimum have this:
```[/Script/Pal.PalGameWorldSettings]
OptionSettings=()```
with the setting inside the parentheses which did not account for.

Twinki14:
Probably shouldn't change `PORT` to `GAME_PORT`, that change will break any current configurations, maybe wouldn't be a problem for a major version bump, but considering how actively developed & released this docker image at the moment, I'd lean towards leaving it as `PORT` or allowing both `PORT` & `GAME_PORT` to exist, and adding a deprecation warning in the readme

#40 Is also brewing up some potential ways to properly support such a setting, without need to manually touch `PalServerSettings.ini`, I really don't like the idea of the dockerfile touching my `PalServerSettings.ini` at all since it could corrupt any changes made outside the container, like a persistent volume

xHyperElectric:
> #40 Is also brewing up some potential ways to properly support such a setting, without need to manually touch `PalServerSettings.ini`, I really don't like the idea of the dockerfile touching my `PalServerSettings.ini` at all since it could corrupt any changes made outside the container, like a persistent volume

I see that, this is exciting

thijsvanloef:
> Probably shouldn't change `PORT` to `GAME_PORT`, that change will break any current configurations, maybe wouldn't be a problem for a major version bump, but considering how actively developed & released this docker image at the moment, I'd lean towards leaving it as `PORT` or allowing both `PORT` & `GAME_PORT` to exist, and adding a deprecation warning in the readme about `PORT`

@Twinki14 You are absolutly correct, although I agree that GAME_PORT is a better name, I can't change it currently as it would break running servers.

thijsvanloef:
Closing this MR, resolved in https://github.com/thijsvanloef/palworld-server-docker/pull/54

Issue #63, opened 2024-01-23
Permission issues running on Kubernetes


Slyke:
## Describe the bug

Server fails to install on first run due to not being able to write to filesystem.

## To Reproduce

Deploy as a Kubernetes pod.

## Expected behavior

Server should start up

## Container/Host Logs
```
*****EXECUTING USERMOD*****
usermod: no changes
*****STARTING INSTALL/UPDATE*****
tid(22) burning pthread_key_t == 0 so we never use it
Redirecting stderr to '/home/steam/Steam/logs/stderr.txt'
Logging directory: '/home/steam/Steam/logs'
[  0%] Checking for available updates...
[----] Verifying installation...
[  0%] Downloading update...
[  0%] Checking for available updates...
[----] Download complete.
[----] Extracting package...
[----] Extracting package...
[----] Extracting package...
[----] Extracting package...
[----] Installing update...
[----] Installing update...
[----] Installing update...
[----] Installing update...
[----] Installing update...
[----] Installing update...
[----] Installing update...
[----] Installing update...
[----] Cleaning up...
[----] Update complete, launching Steamcmd...
tid(34) burning pthread_key_t == 0 so we never use it
Redirecting stderr to '/home/steam/Steam/logs/stderr.txt'
Logging directory: '/home/steam/Steam/logs'
[  0%] Checking for available updates...
[----] Verifying installation...
Steam Console Client (c) Valve Corporation - version 1705108307
-- type 'quit' to exit --
Loading Steam API...OK

Connecting anonymously to Steam Public...OK
Waiting for client config...OK
Waiting for user info...OK
 Update state (0x3) reconfiguring, progress: 0.00 (0 / 0)
 Update state (0x61) downloading, progress: 0.00 (0 / 4729437191)
 Update state (0x61) downloading, progress: 2.29 (108384492 / 4729437191)
 Update state (0x61) downloading, progress: 5.08 (240183440 / 4729437191)
 Update state (0x61) downloading, progress: 7.53 (356275404 / 4729437191)
 Update state (0x61) downloading, progress: 10.33 (488396193 / 4729437191)
 Update state (0x61) downloading, progress: 13.65 (645404772 / 4729437191)
 Update state (0x61) downloading, progress: 16.69 (789113983 / 4729437191)
 Update state (0x61) downloading, progress: 19.91 (941427815 / 4729437191)
 Update state (0x61) downloading, progress: 23.01 (1088228455 / 4729437191)
[...]

PR #65, opened 2024-01-23
fix: added basic yamls for kubernetes support.

beneiltis:
Just added support for kubernetes. We can use this as a starting point. I run it like this in my cluster and it works like a charm. Thx for the nice work @thijsvanloef

### DIFF
--- a/k8s/deployment.yaml
+++ b/k8s/deployment.yaml
@@ -4,7 +4,6 @@ metadata:
   labels:
     app: palworld-server
   name: palworld-server
-  namespace: default
 spec:
   replicas: 1
   selector:
@@ -16,38 +15,28 @@ spec:
         app: palworld-server
     spec:
       containers:
-      - env:
-        - name: "PLAYERS"
-          value: "16"
-        - name: "PORT"
-          value: "8211"
-        - name: "MULTITHREADING"
-          value: "true"
-        - name: "COMMUNITY"
-          value: "true"
-        - name: "SERVER_NAME"
-          value: "YOUR SERVER NAME"
-        - name: "ADMIN_PASSWORD"
-          value: "YOURPASSWORD"
-        - name: "UPDATE_ON_BOOT"
-          value: "true"
-        image: thijsvanloef/palworld-server-docker
-        imagePullPolicy: IfNotPresent
-        name: palworld-server
-        ports:
-        - containerPort: 8211
-          name: "8211-palworld"
-          protocol: UDP
-        - containerPort: 27015
-          name: "27015-palworld"
-          protocol: UDP
-        resources:
-          requests:
-            cpu: 500m
-            memory: 512Mi
-        volumeMounts:
-        - mountPath: /palworld
-          name: datadir
+        - name: palworld-server
+          image: thijsvanloef/palworld-server-docker
+          imagePullPolicy: IfNotPresent
+          ports:
+            - containerPort: 8211
+              name: 8211-palworld
+              protocol: UDP
+            - containerPort: 27015
+              name: 27015-palworld
+              protocol: UDP
+          env:
+            - name: ADMIN_PASSWORD
+              valueFrom:
+                secretKeyRef:
+                  name: palworld-secrets
+                  key: rconPassword
+          envFrom:
+            - configMapRef:
+                name: palworld-cm
+          volumeMounts:
+            - mountPath: /palworld
+              name: datadir
       volumes:
       - name: datadir
         persistentVolumeClaim:
--- a/k8s/deployment.yaml
+++ b/k8s/deployment.yaml
@@ -1,0 +1,54 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  labels:
+    app: palworld-server
+  name: palworld-server
+  namespace: default
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: palworld-server
+  template:
+    metadata:
+      labels:
+        app: palworld-server
+    spec:
+      containers:
+      - env:
+        - name: "PLAYERS"
+          value: "16"
+        - name: "PORT"
+          value: "8211"
+        - name: "MULTITHREADING"
+          value: "true"
+        - name: "COMMUNITY"
+          value: "true"
+        - name: "SERVER_NAME"
+          value: "YOUR SERVER NAME"
+        - name: "ADMIN_PASSWORD"
+          value: "YOURPASSWORD"
+        - name: "UPDATE_ON_BOOT"
+          value: "true"
+        image: thijsvanloef/palworld-server-docker
+        imagePullPolicy: IfNotPresent
+        name: palworld-server
+        ports:
+        - containerPort: 8211
+          name: "8211-palworld"
+          protocol: UDP
+        - containerPort: 27015
+          name: "27015-palworld"
+          protocol: UDP
+        resources:
+          requests:
+            cpu: 500m
+            memory: 512Mi
+        volumeMounts:
+        - mountPath: /palworld
+          name: datadir
+      volumes:
+      - name: datadir
+        persistentVolumeClaim:
+          claimName: palworld-server-datadir
--- a/k8s/pvc.yaml
+++ b/k8s/pvc.yaml
@@ -1,0 +1,15 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  labels:
+    app: palworld-server
+  name: palworld-server-datadir
+  namespace: default
+spec:
+  accessModes:
+  - ReadWriteOnce
+  resources:
+    requests:
+      storage: 10Gi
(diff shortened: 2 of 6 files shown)


k8s/deployment.yaml
```diff
@@ -0,0 +1,54 @@
...
+  selector:
+    matchLabels:
+      app: palworld-server
+  template:
+    metadata:
+      labels:
+        app: palworld-server
+    spec:
+      containers:
+      - env:
```
Slyke, line 19: Maybe do environment variables like this:
```
        - name: ADMIN_PASSWORD
          valueFrom:
            secretKeyRef:
              name: palworld-secrets
              key: rconPassword
        envFrom:
        - configMapRef:
            name: palworld-cm
```

secrets.palworld.yml:
```
apiVersion: v1
kind: Secret
metadata:
  name: palworld-secrets
type: Opaque
stringData:
  rconPassword: yourRconPassword
```

configmap-palworld.yml
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: palworld-cm
data:
  PUID: "1000"
  PGID: "1000"
  PORT: "8211" # Optional but recommended
  PLAYERS: "16" # Optional but recommended
  MULTITHREADING: "false"
  RCON_ENABLED: "true"
  RCON_PORT: "25575"
  COMMUNITY: "false"  # Enable this if you want your server to show up in the community servers tab, USE WITH SERVER_PASSWORD!
  # Enable the environment variables below if you have COMMUNITY=true
  # SERVER_PASSWORD: "yourServerPassword"
  SERVER_NAME: "Server_Name" # Note, it seems spaces are not supported, so use underlines.
```

Slyke:
> @Slyke are you able to confirm that this fixes #63?

Will check tonight after work.

k8s/pvc.yaml
```diff
@@ -0,0 +1,15 @@
...
+    app: palworld-server
+  name: palworld-server-datadir
+  namespace: default
+spec:
+  accessModes:
+  - ReadWriteOnce
+  resources:
+    requests:
+      storage: 10Gi
+  storageClassName: local-path
```
cdrage, line 14: storageClassName and volumeMode shouldn't be specified with `local-path`. The k8s server has default storage providers that can automatically provision it. 

(I use local-path too with k3s / k0s too.. haha).

volumeMode

k8s/pvc.yaml
```diff
@@ -0,0 +1,15 @@
...
+  volumeMode: Filesystem
```
cdrage, line 15: not required as k8s default is Filesystem anyways

k8s/deployment.yaml
```diff
@@ -0,0 +1,54 @@
...
+        - name: "ADMIN_PASSWORD"
+          value: "YOURPASSWORD"
+        - name: "UPDATE_ON_BOOT"
+          value: "true"
+        image: thijsvanloef/palworld-server-docker
+        imagePullPolicy: IfNotPresent
+        name: palworld-server
+        ports:
+        - containerPort: 8211
+          name: "8211-palworld"
```
cdrage, line 39: no need quotes

k8s/deployment.yaml
```diff
@@ -0,0 +1,54 @@
...
+          protocol: UDP
+        - containerPort: 27015
+          name: "27015-palworld"
```
cdrage, line 42: no need quotes

k8s/deployment.yaml
```diff
@@ -0,0 +1,54 @@
...
+          protocol: UDP
+        resources:
+          requests:
+            cpu: 500m
```
cdrage, line 46: should remove..

k8s/deployment.yaml
```diff
@@ -0,0 +1,54 @@
...
+        imagePullPolicy: IfNotPresent
+        name: palworld-server
+        ports:
+        - containerPort: 8211
+          name: "8211-palworld"
+          protocol: UDP
+        - containerPort: 27015
+          name: "27015-palworld"
+          protocol: UDP
+        resources:
```
cdrage, line 44: should remove entire resources section IMHO

k8s/service.yaml
```diff
@@ -0,0 +1,24 @@
...
+spec:
+  externalTrafficPolicy: Cluster
+  internalTrafficPolicy: Cluster
+  ipFamilies:
+  - IPv4
+  ports:
+  - name: palworld1
+    port: 8211
+    protocol: UDP
+    targetPort: 8211
```
cdrage, line 17: targetPort should be `8211-palworld`

k8s/service.yaml
```diff
@@ -0,0 +1,24 @@
...
+  - name: palworld2
+    port: 27015
+    protocol: UDP
+    targetPort: 27015
```
cdrage, line 21: should be `27015-palworld`

k8s/service.yaml
```diff
@@ -0,0 +1,24 @@
...
+kind: Service
+metadata:
+  labels:
+    app: palworld-server
+  name: palworld-server
+  namespace: default
+spec:
+  externalTrafficPolicy: Cluster
+  internalTrafficPolicy: Cluster
+  ipFamilies:
```
cdrage, line 11: not needed

k8s/service.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    app: palworld-server
+  name: palworld-server
+  namespace: default
+spec:
+  externalTrafficPolicy: Cluster
+  internalTrafficPolicy: Cluster
```
cdrage, line 10: not needed

k8s/service.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    app: palworld-server
+  name: palworld-server
+  namespace: default
+spec:
+  externalTrafficPolicy: Cluster
```
cdrage, line 9: not needed

cdrage (changes requested):

k8s/service.yaml
Slyke: `targetPort` isn't needed if `port` is the same, correct?

cdrage: targetport is what is being mapped (service will expose 8211), it's good to point to the port 'name' that's on the deployment (since it was named 8211-palworld)

Slyke:
@thijsvanloef It seems to be working without my initContainer updating the permissions. Whether it's due to just needing a one time update or not is unknown. I would say that it's safe to merge though.

Slyke (approved):

k8s/pvc.yaml
alex4108: good point, I'm running longhorn in my homelab so local-path would kill my usage.

Slyke (approved):

beneiltis:
Thx people :-) I have changed everything as requested. Realy good suggestions you guys made :-) much appreciated.
I have checked it with my setup and it is still working great. Ready for merge :-)

elonmir:
Have you thought about usage of `kustomization`, then you could roll it out with a single apply command.

elonmir:
Just thinking out loud, maybe also adding a cronjob resource for backing stuff up? Out of scope?

beneiltis:
I would suggest we finish this PR and iterate from there on. Otherwise this is gonne be a never ending PR ;-)

thijsvanloef:
Nice work all 💪! @beneiltis should I wait before I merge this as @cdrage and @alex4108 are still pending reviewers, or are you confident?

beneiltis:
@thijsvanloef you can merge it from my point of view. I have modified all parts according to the guys suggestions. I also tried all manifests on my own server. everything works great.

elonmir:
> I would suggest we finish this PR and iterate from there on. Otherwise this is gonne be a never ending PR ;-)

Yes, sir! =) Good job anyways.

k8s/secret.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: palworld-secrets
+type: Opaque
+stringData:
+  rconPassword: yourRconPassword
```
alex4108, line 7: "yourRconPassword in base64 format" ?

thijsvanloef:
Awesome, once again thank you for your work @beneiltis!

alex4108 (approved):

k8s/secret.yaml
Slyke: No, it's plaintext in the manifest files. It's base64 if you `kubectl describe` though.

I usually call it something like `example.secret.yaml`, make a copy, put the actual password in, apply the yaml with `kubectl apply -f secrets.yaml` then `rm secrets.yaml`.

cdrage:
Confirm it works my side too, awesome work @beneiltis  :)

Issue #62, opened 2024-01-23
Stopping does not appear to be graceful


fryfrog:
When I use `docker stop palworld` to stop the container, it doesn't seem to gracefully shutdown the server. Instead, it waits until the end of the timeout and just kills it. The default is 10 seconds, I've tried 30 and 60 and they all just go until the end and kill it.

In the container I've tried sending it `15` and `9` via `kill`, but they don't seem to be particularly graceful either.

There does appear to be an rcon / admin command `/Shutdown {Seconds} {MessageText}` ([source](https://nodecraft.com/support/games/palworld/palworld-server-admin-commands), so maybe that could be incorporated somehow?

I'll do more research and see what I can figure out, but maybe someone has better ideas?

fryfrog:
I am weird and use systemd service files to `docker run`, so I've settled on installing `rcon-cli` and setting `ExecStop=/usr/bin/rcon -a 127.0.0.1:25575 -p mypassword "Shutdown 10"` to get graceful shutdowns for now.

PR #98, opened 2024-01-25
Graceful stop

Dashboy1998:
## Context 

* Add in graceful stop

## Choices

* Using rcon was mentioned by @thijsvanloef in #62
* In the event rcon is not used then kill is used instead

## Test instructions

1. Run with `RCON_ENABLED=true`
2. Join the game and move to a different location in game
3. Stop the container with ctrl + c
4. Restart container and join game again to verify it saved where you ended
5. Repeat 1 - 3 with `RCON_ENABLED=false` NOTE: this does not save without rcon


thijsvanloef:
Thank you for the PR! I will check it once I get home from work :D

Issue #96, opened 2024-01-24
Support migrate from existing data


Ericwww:
Thank you for providing the docker image and compose file.

I migrated my existing server to use docker deployment today.
Recovering from existing data requires some manual operations. I think this operation can be integrated into the docker image to facilitate users to migrate servers.

I saw a [migration solution](https://bbs.nga.cn/read.php?tid=39075608&rand=541) from a game forum in China like this:

```shell
cp -r oldServer:steamapps/common/PalServer/Pal/Saved/SaveGames/0/{SERVER_NAME}/ newServer:steamapps/common/PalServer/Pal/Saved/SaveGames/0/{SERVER_NAME}/
```

And then 

```shell
sed -i "s/DedicatedServerName=[A-Z0-9]*/DedicatedServerName={SERVER_NAME}/" /palworld/Pal/Saved/Config/LinuxServer/GameUserSettings.ini
```

Now I'm not sure if the `GameUserSettings.ini` file is generated before starting the game server. 
If so this problem would be much simpler. 
If not, I have no idea how to solve this problem.

I can't test it because I'm with Apple Silicon.

If anyone has ideas, welcome to discuss.

Jortana:
I used to run the server directly on Windows, but later I migrated it to a Docker environment.

My approach was to first enter the new server and create a character, in order to let my new server generate some necessary files. 

Then, I would change the DedicatedServerName in the `GameUserSettings.ini` file of the new server. Also, I deleted the `steamapps/common/PalServer/Pal/Saved/SaveGames/0/{SERVER_NAME}/` folder in the new server and replace it with the folder from the old server.

By doing this and restarting the server, All players could play normally.

PR #114, opened 2024-01-25
Support migrate from an existing server

Ericwww:
## Context 

To solve #96 

I think `/palworld/Pal/Saved/Config/LinuxServer/GameUserSettings.ini` is generated after the service is started for the first time, so I can only write an additional script to facilitate migration.

After several experiments, the migration plan was determined to be successful.

## Test instructions

This script is also tested.

image

image

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,6 +176,29 @@ For a more detailed list of explanations of server settings go to: [shockbyte](h
 > If the `<mount_folder>/Pal/Saved/Config/LinuxServer/PalWorldSettings.ini` is empty,
 > delete the file and restart the server, a new file with content will be created.
 
+## Migrate From Existing Server
+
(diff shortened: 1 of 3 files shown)


README.md
thijsvanloef, line 1: I prefer to have this in a seperate readme in the migration folder

thijsvanloef (changes requested):

README.md
Ericwww: OK, separated it.

thijsvanloef (changes requested):
Could you also possibly document the way to do it manually in the readme?

RoryKirsi:
Is any option to work with the Windows server migrate to Linux server? (lock steam_appid.txt)

PR #123, opened 2024-01-26
Increase recommended disk size to 12GB.

wli:
The steamcmd update requires an extra 2GB of free disk space. The container seems to be using around 8.7GB, which results in a steam update failure of `Error! App '2394010' state is 0x226 after update job.` since there isn't enough space to grab the new version in a 10GB disk. Bumping this to 12GB should result in ~3.4GB of free space, so there is a small buffer for future update size increases.

## Context 

* Update documentation to help others who want to set up the container correctly to begin with.

## Test instructions

1. N/A - documentation update only

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ This Docker container has been tested and will work on both Linux (Ubuntu/Debian
 |----------|---------|------------------------------------------|
 | CPU      | 4 cores | 4+ cores                                 |
 | RAM      | 16GB    | Recommend over 32GB for stable operation |
-| Storage  | 4GB     | 10GB                                     |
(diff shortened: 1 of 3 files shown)


MSpreckels:
Hey, maybe change that in the [helm chart](https://github.com/thijsvanloef/palworld-server-docker/blob/9518170152528f7ec617a91a52bd8b91d93ef610/chart/values.yaml#L27) and [kubernetes PVC](https://github.com/thijsvanloef/palworld-server-docker/blob/9518170152528f7ec617a91a52bd8b91d93ef610/k8s/pvc.yaml#L12), aswell?

PR #136, opened 2024-01-26
Add a Korean-translated README.md

seoseonyu:
## Context 

- Added a Korean translated README.md for Korean users.
- For another language translation, I created a /docs folder and created a translation file in it.
- Added links to jump to language-specific documentation in each README.md file.

## Choices

- We've created a translation document so that Korean users don't suffer from incorrect translations using Google Translate.

## Test instructions

1. 


thijsvanloef:
@seoseonyu Thank you for taking the time to translate the Readme! I did have 2 questions:

- The current README is quite often being changed, how would I change the korean version?
- Is it possible to name the folder `./docs/kr` to reflect the korean TLD?

seoseonyu:
First of all, thank you for managing that repository.

For the first question, I personally recommend using a combination of two methods.
1. subscribe to changes in the original via github watch to reflect changes as they happen.
2. assign a version number to the README.md to compare the current state of the original and the translation, and update it if the translation has a lower version.

Change './docs/en` -> './docs/kr` to reflect the Korean TLD.

If you have any additional suggestions, please let me know!

README.md
```diff
@@ -10,6 +10,8 @@
 
 [Chat with the community on Discord](https://discord.gg/UxBxStPAAE)
 
+[English](/README.md) | [한국어](/docs/ko/README.md)
```
thijsvanloef, line 13: don't forget to change the path :P

thijsvanloef (changes requested):

README.md
seoseonyu: Ah! I missed something important, thanks.

thijsvanloef (approved):

PR #141, opened 2024-01-26
modify deployment and chart for opionated deployment

Mercwri:
## Context 

It is relatively unsafe with the way Palworld handles it's server terminations to leave the default deployment setting of RollingUpdate. Switching to Recreate ensures the old pod is terminated and not writting to the volume while the new pod is starting.

## Choices

Opionated change is done in the suggested deployment yaml and the chart sets Recreate by default but leave it a configurable option.

## Test instructions

Deployed to my local cluster, valiadate that deployment still works and that it terminates the pod prior to scheduling the new pod.

### DIFF
--- a/chart/templates/deployments.yaml
+++ b/chart/templates/deployments.yaml
@@ -8,9 +8,9 @@ spec:
   selector:
     matchLabels:
       app.kubernetes.io/component: "{{ .Release.Name }}-server"
+  strategy:
+    type: "{{ .Values.server.strategy }}"
   template:
-    strategy:
-      type: "{{ .Values.server.strategy }}"
     metadata:
       annotations:
         {{- with .Values.server.annotations }}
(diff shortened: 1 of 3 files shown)


thijsvanloef:
Hi @Mercwri Thanks for taking the time to create this PR, i've marked it "help wanted" as I am currently not able to test it myself

chart/templates/deployments.yaml
Filipe-Souza, line 1: Move the strategy property up to spec:

```

  selector:
    matchLabels:
      app.kubernetes.io/component: "{{ .Release.Name }}-server"
  strategy:
    type: "{{ .Values.server.strategy }}"
  template:
    metadata:
      annotations:
 ```

Filipe-Souza (changes requested):
After moving the property OK
Screenshot 2024-01-26 224618

chart/templates/deployments.yaml
Mercwri: Good catch, just pushed the fix.

Filipe-Souza (approved):

PR #145, opened 2024-01-26
Added fix for blank PalWorld Settings file

Dashboy1998:
## Context 

* Removes the need to manually delete the PalWorldSettings.ini file if it is blank to trigger a copy of the default file at next startup.

## Choices

* Using grep since that is already included in the image.

## Test instructions

1. Start the container without `:/palworld/Pal/Saved/Config/LinuxServer/PalWorldSettings.ini`
2. Verify the file is created
3. Start the container with an empty (zero bytes, you can use touch) `:/palworld/Pal/Saved/Config/LinuxServer/PalWorldSettings.ini`
4. Verify the file is overwritten
5. Start container with a valid `:/palworld/Pal/Saved/Config/LinuxServer/PalWorldSettings.ini` (Ideally modify a value)
6. Verify the file not overwritten

### DIFF
--- a/README.md
+++ b/README.md
@@ -202,9 +202,6 @@ Please keep in mind that the ENV variables will always overwrite the changes mad
 
 For a more detailed list of explanations of server settings go to: [shockbyte](https://shockbyte.com/billing/knowledgebase/1189/How-to-Configure-your-Palworld-server.html)
 
-> [!TIP]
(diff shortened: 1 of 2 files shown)


thijsvanloef:
I might have set shellcheck too strict, for now you can use:

`# shellcheck disable=SC2143` 

In the line above the "faulty" line

Ggust498:
Where do i get a valid PalworldSettings.ini so i can overwrite mine?

PR #154, opened 2024-01-27
Updated links

Dashboy1998:
## Context 

* Current link prompts you to login if you are not.
* If you are logged in but have not verified your email it will require you to do so.
* If you are logged in and have verified your email then it will forward you to the new link.

## Choices

* Changed to be the link to work without an account.

## Test instructions

1. Go to link / click images while not logged in.


Dashboy1998:
I'll say this is a strange bug. I only noticed it because I haven't had the need to make a docker hub account until now.

PR #155, opened 2024-01-27
Added setting graceful stop period

Dashboy1998:
## Context 

* Sometimes the server takes longer than the default 10 seconds to shutdown #156 

## Choices

* Added settings to set the grace period

## Test instructions

* Unsure of the maximum amount of time a server might take to shutdown.
* Unsure how create a scenario which makes the container take longer than 10 seconds to stop.
1. Ideally attempt to stop a which takes longer than 10 seconds to shutdown.


thijsvanloef:
What is your reasoning behind the 30s? And have you observed that within those 30s the container always shuts down gracefully?
I aks this because I was not able to test with a server that takes longer then 10 seconds to shutdown gracefully

Dashboy1998:
> What is your reasoning behind the 30s? And have you observed that within those 30s the container always shuts down gracefully? I aks this because I was not able to test with a server that takes longer then 10 seconds to shutdown gracefully

So far the highest I've seen it take is 24 seconds.

thijsvanloef (approved):

PR #167, opened 2024-01-28
Minimize Root Processes

Dashboy1998:
## Context 

* Currently you start the container as root and during the start up process everything is ran as root except the command with `su`.
image

* Currently when backups are created they are created and owned by root however on next boot the permission is changed to steam:steam

* When starting with a PUID or PGID that is not 1000 you get the following error: `/tmp/dumps is not owned by us - delete and recreate`
```
palworld-server-test  | *****STARTING INSTALL/UPDATE*****
palworld-server-test  | tid(22) burning pthread_key_t == 0 so we never use it
palworld-server-test  | Redirecting stderr to '/home/steam/Steam/logs/stderr.txt'
palworld-server-test  | Logging directory: '/home/steam/Steam/logs'
palworld-server-test  | /tmp/dumps is not owned by us - delete and recreate
palworld-server-test  | [  0%] Checking for available updates..
```

* Currently When starting with a PGID that is not 1000 then /home/steam/ has the group 1000 instead of steam.

## Choices

* Overall trying to reduce having to run as root. 
* Moved all commands which needed `su` to start.sh
* Other commands in start.sh should be ran as steam anyways
* Everything in start.sh should only need to be ran as steam user

## Test instructions

**Unless explicitly stated your PUID or PGID should not be 0 for testing**

image
1. Start container with a PGID that is neither 0 or 1000
    1.  verify you do not get this error: `/tmp/dumps is not owned by us - delete and recreate`
    2.  verify /home/steam is owned by steam:steam: `docker exec palworld-server bash -c "ls -ll /home/steam"`
2. After the server has finished booting use top to verify the only processes running as root are init.sh, su, and top: `docker exec -t palworld-server top`
3. Run a backup as root and verify that the file is owned by steam:steam: `docker exec palworld-server bash -c "backup && ls -l /palworld/backups/"`
4. Run a backup as steam and verify that the file is owned by steam:steam: `docker exec -u steam palworld-server bash -c "backup && ls -l /palworld/backups/"`
5. Test rcon-cli as root: `docker exec -it palworld-server-test rcon-cli`
6. Test rcon-cli as steam: `docker exec -it -u steam palworld-server-test rcon-cli`
7. Verify you don't get this message `/tmp/dumps is not owned by us - delete and recreate` when:
    1. Start container with a PGID that is 0
    2. Start container with a PUID that is 0
    3. Start container with a PUID & PGID that is 0
8. Verify you still get this message `Running as root is not supported, please fix your PUID and PGID!` when:
    1. Start container with a PGID that is 0
    2. Start container with a PUID that is 0
    3. Start container with a PUID & PGID that is 0

### DIFF
--- a/scripts/start.sh
+++ b/scripts/start.sh
@@ -85,6 +85,6 @@ default:
 EOL
 
 printf "\e[0;32m*****STARTING SERVER*****\e[0m\n"
-echo "bash -c '${STARTCOMMAND[@]}'"
+echo "bash -c '${STARTCOMMAND[*]}'"
 "${STARTCOMMAND[@]}"
 
--- a/scripts/start.sh
+++ b/scripts/start.sh
@@ -61,7 +61,7 @@ if [ ! "$(grep -s '[^[:space:]]' /palworld/Pal/Saved/Config/LinuxServer/PalWorld
     printf "\e[0;32m*****GENERATING CONFIG*****\e[0m\n"
 
     # Server will generate all ini files after first run.
-    su steam -c "timeout --preserve-status 15s ./PalServer.sh 1> /dev/null "
+    timeout --preserve-status 15s ./PalServer.sh 1> /dev/null
 
     # Wait for shutdown
     sleep 5
--- a/scripts/start.sh
+++ b/scripts/start.sh
@@ -1,5 +1,10 @@
 #!/bin/bash
 
+if [ "${UPDATE_ON_BOOT}" = true ]; then
+    printf "\e[0;32m*****STARTING INSTALL/UPDATE*****\e[0m\n"
+    /home/steam/steamcmd/steamcmd.sh +force_install_dir "/palworld" +login anonymous +app_update 2394010 validate +quit
+fi
+
 STARTCOMMAND=("./PalServer.sh")
 
 if [ -n "${PORT}" ]; then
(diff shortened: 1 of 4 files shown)


scripts/start.sh
```diff
@@ -80,6 +85,6 @@ default:
 EOL
 
 printf "\e[0;32m*****STARTING SERVER*****\e[0m\n"
-echo "bash -c '${STARTCOMMAND[*]}'"
-su steam -c "bash -c '${STARTCOMMAND[*]}'"
+echo "bash -c '${STARTCOMMAND[@]}'"
+"${STARTCOMMAND[@]}"
```
Dashboy1998, line 89: Had to change `"${STARTCOMMAND[*]}"` to `"${STARTCOMMAND[@]}"` otherwise you get the following error: `./start.sh: line 89: ./PalServer.sh -port=8211 -players=16 -servername="World of Pals" -serverdescription="" -serverpassword="worldofpals" -adminpassword="adminPasswordHere" -queryport=27015 -useperfthreads -NoAsyncLoadingThread -UseMultithreadForDS: No such file or directory`

Changed the echo just to match the command.

thijsvanloef (approved):
I've tested it, and everything works fine, however this is a big PR that I would also like a second opinion on. Posting this in the discord, hopefully someone else would be able to look at it.

Once again thank you for all the contributions :D

MSpreckels:
LGTM

Maybe eventually we can go for non root when we run steam non root (if that is even possible)

PR #180, opened 2024-01-28
Single sed

Dashboy1998:
## Context 

* Combines all sed command that modify `/palworld/Pal/Saved/Config/LinuxServer/PalWorldSettings.ini` into one command

## Choices

* Using regular expressions with back referencing so if the environmental variable is not set then value is not replace (Although technically it's replaces with itself)

## Test instructions

1. Try starting the container with any valid configuration of environmental variables set.

### DIFF
--- a/scripts/start.sh
+++ b/scripts/start.sh
@@ -295,7 +295,7 @@ RCON_PORT="${RCON_PORT:-\1}"
 # Requires all non number values to be declared as those expressions need updating to use a replacement
 sed -E  \
     -e "s/ServerPlayerMaxNum=([0-9]*)/ServerPlayerMaxNum=$PLAYERS/" \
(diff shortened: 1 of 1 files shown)


thijsvanloef:
While this is a "cleaner" way of doing it, I'm not fully convinced this is a solution I want to merge.
Mainly because it adds a lot of complexity to the code, making it less readable.

Also if I make a mistake anywhere in the sed command, the whole command will fail instead of only one setting.

I need to think a little more about this one

Dashboy1998:
> While this is a "cleaner" way of doing it, I'm not fully convinced this is a solution I want to merge. Mainly because it adds a lot of complexity to the code, making it less readable.
> 
> Also if I make a mistake anywhere in the sed command, the whole command will fail instead of only one setting.
> 
> I need to think a little more about this one

While it does make it slightly more difficult to understand also means you only write to the file once.

thijsvanloef:
Decided against implementing this change

PR #201, opened 2024-01-29
- Update the way env variables are managed

Filipe-Souza:
## Context 

* The easer option is to give the user a "config file", in this case a .env.example file and get it to be modified;
* The goal is avoid the hassle with copy/pasting commands with wrong formatting, different values and different OS;

## Choices

* Environment variables files are a common practice in programming environment, it's easier on users to manage and find what they need, and provide us a source of truth to solve problems;

## Test instructions

1. Follow up the README instructions to how to set-up with docker run and docker-compose;


thijsvanloef (changes requested):
While I agree that this makes it a lot easier, my though process is that the majority does not clone this entire repo but instead just pulls the image from docker hub.

I would like to include the .env.example in the repo, but not change the docker-compose.yml examples. But instead create a disclaimer in the "How to deploy - Docker Compose" heading saying that this can also be managed using an .env file

Filipe-Souza:
Makes sense, I'm going to update the PR

PR #206, opened 2024-01-29
Add Daily Reboot Cronjob (K8s)

NotaSF:
Requires the k8s deployment to be in the **palworld** namespace.

Currently, it is set to run at 9:30am UTC daily.

I'm newer to configuring such things, and will work to improve/cleanup a little bit. :)

## Context 

* This launches a cronjob to run a shell script (defined in cm.yaml), which sends a broadcast notifying players that the server is saving. After 30 seconds, it does a formal backup using the established 'exec pod/podnamehere backup' command in the repo. After 30 seconds later, it sends a broadcast notifying players that the server will reboot in 60 seconds; 30 seconds; then each second for the last 10 seconds. It the runs the "Shutdown 1" command via rcon-cli, waits 30 additional seconds, then redeploys the application.

## Choices

* I prefer the K8s cluster schedule and perform a cronjob so I can keep track of it and also rollback if needed.
* This made the most sense to me, as it does not intrude or alter the main image itself.
* I wanted to learn more about how CronJobs work in Kubernetes.

## Test instructions

1. Copy the folder and its contents
2. Edit the cronjob.yaml schedule to '* * * * *' so it schedules it immediately.
3. Log into server
4. Apply the yaml files, specifically applying cronjob.yaml last.
5. Wait for the broadcasts, also monitor the logs of the job created using `kubectl -n palworld logs job/jobnamehere`
6. Server should disconnect once timer reaches 0; redeployment should be under way.
7. Server should come back up once redeployment is finished.


Twinki14:
I would say,

- Add this to the helm chart, and make it opt-in, it's far better suited for the helm chart
- If we still want this in the repository, let's store it only as an example, perhaps in a directory `k8s/examples/daily-restart`, with an included readme to explain it and note any requirements/changes someone would need to make to use it

Twinki14 (changes requested):
I would say,

- Add this to the helm chart, and make it opt-in, it's far better suited for the helm chart
- If we still want this in the repository (outside the helm chart), let's store it only as an example, perhaps in a directory `k8s/examples/daily-restart`, with an included readme to explain it and note any requirements/changes someone would need to make to use it

Really, since we have a helm chart now, we don't need to have the raw k8s manifests like we do now anyway. Helm is extremely simple to setup in a k8s environment, and is the defacto standard in devops

k8s/cronjob-reboot/cm.yaml
```diff
@@ -0,0 +1,43 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: restart-deployment-configmap
+  namespace:  palworld
```
MSpreckels, line 5: please remove the namespace

k8s/cronjob-reboot/cm.yaml
```diff
@@ -0,0 +1,43 @@
...
+    sleep 1
+    kubectl exec -n palworld -i pod/$cont rcon-cli "Broadcast Rebooting_in_3_seconds..."
+    sleep 1
+    kubectl exec -n palworld -i pod/$cont rcon-cli "Broadcast Rebooting_in_2_seconds..."
+    sleep 1
+    kubectl exec -n palworld -i pod/$cont rcon-cli "Broadcast Rebooting_in_1_seconds..."
+    sleep 1
+    kubectl exec -n palworld -i pod/$cont rcon-cli "Shutdown 1"
+    sleep 30
+    kubectl -n palworld rollout restart deployment/palworld-server
```
MSpreckels, line 42: can you make a for loop out of that? also set the namespace via env or as a variable on top like cont, it is way less error prone

k8s/cronjob-reboot/cronjob.yaml
```diff
@@ -0,0 +1,35 @@
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: restart-deployment
+  namespace:  palworld
```
MSpreckels, line 5: please remove the namespace

k8s/cronjob-reboot/cronjob.yaml
```diff
@@ -0,0 +1,35 @@
...
+        spec:
+          serviceAccountName: restart-deploy
+          restartPolicy:  Never
+          containers:
+            - name: kubectl
+              image:  bitnami/kubectl
+              command:
+              - /bin/sh
+              - -c
+              - /restart-script/restart-deployment.sh
```
MSpreckels, line 23: do you even need a CM with the shell script? what about inline?

k8s/cronjob-reboot/role.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+    name: restart-deployment
+    namespace: palworld
```
MSpreckels, line 5: please remove the namespace

k8s/cronjob-reboot/rolebinding.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: restart-deployment
+  namespace: palworld
```
MSpreckels, line 5: please remove the namespace

k8s/cronjob-reboot/serviceaccount.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+      name: restart-deploy
+      namespace: palworld
```
MSpreckels, line 5: please remove the namespace

also: the indentation looks off

MSpreckels (changes requested):
I've reviewed it but as @Twinki14 said, this would fit in the helm chart too. Can you add that?

If you add a kustomization.yaml you could easily patch every namespace as well as the SA reference in the role binding :)

NotaSF:
> I would say,
> 
> * Add this to the helm chart, and make it opt-in, it's far better suited for the helm chart
> * If we still want this in the repository (outside the helm chart), let's store it only as an example, perhaps in a directory `k8s/examples/daily-restart`, with an included readme to explain it and note any requirements/changes someone would need to make to use it
> 
> Really, since we have a helm chart now, we don't need to have the raw k8s manifests like we do now anyway. Helm is extremely simple to setup in a k8s environment, and is basically the defacto standard in any k8s environment

I have made some changes and added helm chart values. Tested on my dev cluster... let me know if some additional changes would be needed.

k8s/cronjob-reboot/cronjob.yaml
NotaSF: That is a good point, I'm not sure actually. Pardon my ignorance, I'm a little new to this - when you say inline, do you mean taking the contents of the shell script and making it a one-liner instead of using a CM and mounting it as a volume?

k8s/cronjob-reboot/cm.yaml
NotaSF: I will sit down and look closer at this for improvement

Twinki14:
> I have made some changes and added helm chart values. Tested on my dev cluster... let me know if some additional changes would be needed.

Nice 👍 My only final nit-pick is the config map script, I think that can definitely be cleaned up with some type of for-loop

k8s/cronjob-reboot/cronjob.yaml
MSpreckels: Hi sorry for the confusion. What i mean is instead of having a configmap mounted you could also put that script in that command as an arg like

```
command:
  - /bin/sh
  - -c
args:
  - |
    cont=$(kubectl -n palworld get pods -o=jsonpath='{.items[?(@.metadata.labels.app=="palworld-server")].metadata.name}')
    kubectl exec -n palworld -i pod/$cont rcon-cli "Broadcast Saving_Server_Data..."
    kubectl exec -n palworld -i pod/$cont rcon-cli save
    sleep 30
    kubectl exec -n palworld -i pod/$cont rcon-cli "Broadcast Backing_up_Server_Data..."
    kubectl exec -n palworld -i pod/$cont backup
 ...
```

k8s/cronjob-reboot/cm.yaml
PaulusParssinen: Following should do the job:

```suggestion
  restart-deployment.sh:  |
    #!/bin/bash
    cont=$(kubectl -n palworld get pods -o=jsonpath='{.items[?(@.metadata.labels.app=="palworld-server")].metadata.name}')

    function exec_rcon_cmd() {
      kubectl exec -n palworld -i pod/$cont rcon-cli "$1"
    }

    exec_rcon_cmd "Broadcast Saving_Server_Data..."
    exec_rcon_cmd save
    sleep 30
    exec_rcon_cmd "Broadcast Backing_up_Server_Data..."
    exec_rcon_cmd backup
    sleep 30

    for i in {30..1}; do
      exec_rcon_cmd "Broadcast Rebooting_in_$i_seconds..."
      sleep 1
    done
    
    exec_rcon_cmd "Shutdown 1"
    sleep 30
    kubectl -n palworld rollout restart deployment/palworld-server
```

MSpreckels: You could also move the foor loop around and have it like 

```
for i in {0..30}
do
  exec_rcon_cmd "Broadcast Rebooting_in_$((30-($i*5)))_seconds..."
  sleep 5
done
```

thijsvanloef:
@NotaSF Is this ready to merge?

NotaSF:
Did some last minute cleanup, moving it to an examples folder instead.

It should be ready for merge, but I must admit I do not know how to update the chart metadata with the newly-added chart values other than manually adding them.

NotaSF:
Found the source of the markdown lint failure - should be good now.

PR #207, opened 2024-01-29
set palworldsettings via envsubst and templating

MSpreckels:
Hi there!

I've got another idea for setting the PalWorldSettings.ini!

In my example I use a template file that has environment variables set for each parameter. Using envsubst I patch those env vars in while also setting a default value for each.

I have tested if the variables are set in file but not ingame, yet.


Twinki14:
This is way cleaner, and way easier to maintain long-term

Some things to test,

- Server name / description, something with spaces, `SERVER_NAME="A server with spaces in it's name"`

MSpreckels:
I wanted to wait for some feedback before i continue working on it and building it further. There is still alot to test about e.g. i already saw that string arent properly parsed. Will continue work on this tomorrow

thijsvanloef:
@MSpreckels Keep doing that you are doing, this sounds like an awesome solution

thijsvanloef:
@MSpreckels keep the naming as close to the PalWorldSettings.ini, except for the ones that introduce breaking changes.

Tonight I'll test your PR as well, thanks for all your work

thijsvanloef:
Also some more things to test: 
- SERVER_NAME=hello world
- SERVER_NAME="hello world"
- SERVER_NAME=https://github.com
- SERVER_NAME="https://github.com"
- SERVER_NAME=special -&\/#@ characters
- SERVER_NAME="special -&\/#@ characters"
Too see if everything gets handled correctly

MSpreckels:
> @MSpreckels keep the naming as close to the PalWorldSettings.ini, except for the ones that introduce breaking changes.
> 
> Tonight I'll test your PR as well, thanks for all your work

will check that tomorrow

> Also some more things to test:
> 
> * SERVER_NAME=hello world
> * "SERVER_NAME=hello world"
> * SERVER_NAME=https://github.com
> * "SERVER_NAME=https://github.com"
> * SERVER_NAME=special -&/#@ characters
> * "SERVER_NAME=special -&/#@ characters"
>   Too see if everything gets handled correctly

These are the results when i put the values in the config map:
1. https://i.imgur.com/ie7T2L0.png
3. https://i.imgur.com/0xZVgkb.png
5. https://i.imgur.com/m2xlytD.png

`SERVER_NAME = value` and `SERVER_NAME = "value"` give the same results

`SERVER_NAME = "\"value\""` breaks it tho

if you could test it using the docker command would be nice

thijsvanloef:
@MSpreckels As for the debug question (sorry missed it completely) I would like a DEBUG env var to switch them on. But disable them by default

vhladiienko:
I don't think env vars work at all on 0.19.1 either 🤔

Twinki14:
> Initial testing does not seem very succesfull, all PalWorldSettings.ini are set correctly, however it does not seem to work ingame: `PalWorldSettings.ini`

This is more likely a problem with the formatting of the ini file, all settings may need to be on one line.

The values are reaching the ini file, which is a really good sign for this method. Now we just need to get the ini formatted correctly so that the server picks up the settings properly.

Twinki14:
> Some considerations:
> 
> * Should i keep the debug output?
> * What kind of naming do i use for envronment variables?

In my opinion,
- Let's always output the environment variables, it's very useful information to have and imo a informational level, not debug level. Outputting the full ini after compilation would also be a bonus
- Take the setting and convert it to snake case, that should be the environment variable naming scheme. `DayTimeSpeedRate` -> `DAY_TIME_SPEED_RATE`, `bIsPvP` -> `B_IS_PVP`

MSpreckels:
Hey thanks for testing! I wonder why it is not formatting correctly. I'm using `tr` to remove all the new lines. Might be a tooling difference between macos and linux? I'll check.

> * Take the setting and convert it to SNAKE_CASE, that should be the environment variable naming scheme.
>   * ...
>   * `bIsPvP` -> `B_IS_PVP`

I'd maybe remove type identifiers like **b** :)

> * Take the setting and convert it to SNAKE_CASE, that should be the environment variable naming scheme.
>   * ...
>   * Prefixing each environment variable with `PALWORLD_SERVER_` might be smart, if we want to support "other" ini sections, but I don't think it's worthwhile

I don't like prefixing them. Some of those variables are already super long which would make it more unreadable.

> * Take the setting and convert it to SNAKE_CASE, that should be the environment variable naming scheme.
>   * ...
>   * This would set us up nicely for the future if we ever wanted to automate the creation of the setting compilation script, we could read settings from a `DefaultPalworldSettings.ini`, and easily convert each setting to a environment variable counterpart

I like that idea!

thijsvanloef:
I've done some initial testing and everything seems to work! Will test more thoroughly tomorrow. Thank you for all the work!

Twinki14:
> I'd maybe remove type identifiers like **b** :)

I only included the modifier with the idea that eventually we'll have an automation script for the compilation script, identifying and dropping type identifiers might add to the logic complexity for such a script, but that is only an eventual idea.

I'm not partial in any direction.

thijsvanloef:
@chkpwd no, but you can test it locally by checking out this branch and running: 
`docker build --tag thijsvanloef/palworld-server-docker:latest`
And then running compose up

Twinki14:
Some final thoughts on this, and these are just my opinions

- After this is merged, the next release should bump from 0.19.1 to 0.50.0, this PR should solidify how we're going to support server settings for probably awhile, and will override anyone's current custom ini's, this will very much be a big breaking change
- If we agree on this, if we wait spend some time waiting before the 0.50.0 release, I can update the helm chart & example docker compose to instead use the "last compatible version" instead of `latest`, remove custom ini support, and remove the k8s manifests
- The readme should also be updated in the wait timeframe

Dashboy1998:
Since we define the default values in the compile-settings if the game updates and changes one of the default values then no one running the image would have the updated default values until we update the compile-settings file and they update their image.

If a value isn't defined the ini file then game seems to use the default value.

Twinki14:
> Since we define the default values in the compile-settings if the game updates and changes one of the default values then no one running the image would have the updated default values until we update the compile-settings file and they update their image.
> 
> If a value isn't defined the ini file then game seems to use the default value.

Hence why an automation script would be very valuable. A scheduled GitHub Action can regularly run, start a Palworld server, copy the `DefaultPalworldSettings.ini`, run an automation script against it, and then output the full compilation script with all default values.

Dashboy1998:
> > Since we define the default values in the compile-settings if the game updates and changes one of the default values then no one running the image would have the updated default values until we update the compile-settings file and they update their image.
> > If a value isn't defined the ini file then game seems to use the default value.
> 
> Hence why an automation script would be very valuable. A scheduled GitHub Action can regularly run, start a Palworld server, copy the `DefaultPalworldSettings.ini`, run an automation script against it, and then output the full compilation script with all default values.

Why not create the compilation script at container start up instead?

Twinki14:
> Why not create the compilation script at container start up instead?

It needs a `DefaultPalworldSettings.ini`, which if it's populated when the server is installed from steamcmd, this might be doable. But I think keeping it outside the container will be better, for documentation purposes

MSpreckels:
> * After this is merged, the next release should bump from 0.19.1 to 0.50.0, this PR should solidify how we're going to support server settings for probably awhile, and will override anyone's current custom ini's, this will very much be a big breaking change

I really dont like that. We are in active early development, changes like that are to be expected. We should test the case when someone is using a save from previous versions and upgrades to a latest image where this change is implemented, so that we know what is going to happen. Maybe we can check for custom values in ini? 
0.20.0 is sufficient IMO.

> If a value isn't defined the ini file then game seems to use the default value.

Thats a good point, somehow I've never tried that.

> Why not create the compilation script at container start up instead?

What do you mean by container start up? Do you mean in the init script? when the docker container is being build?

Dashboy1998:
In the start script read values in default ini to create the config script.

Been trying to find a good tool that let's you modify ini but none seem to support arrays.

MSpreckels:
> In the start script read values in default ini to create the config script.
> 
> Been trying to find a good tool that let's you modify ini but none seem to support arrays.

Thats the thing. Pocket Pair's decision on making a giant array instead of separated values makes this so much harder.

iamspido:
When do you plan to merge this?

thijsvanloef:
@iamspido @MSpreckels Initial testing resulted in no errors, will resume testing tomorrow and probably release tomorrow :)

thijsvanloef:
Did some more testing today, was not able to find any errors, @MSpreckels once merge conflict has been resolved, ready to merge :)

iamspido:
I don't know if that was planned, but the environment variable BASE_CAMP_WORKER_MAXNUM has been changed to BASE_CAMP_WORKER_MAX_NUM.
By setting BASE_CAMP_WORKER_MAX_NUM: "20" it has no effect, and you're still limited to the default of 15.

Issue #38, opened 2024-01-21
check for available updates automatically


GMBridge:
some kind of cron job to check for pending game updates & automatically restart the server to apply them would be nice .

PR #209, opened 2024-01-29
Cron update

Luatan:
## Context 

* adds a Cronjob to check for updates
* Cronjob can be set like the backup cronjob from @Dashboy1998
* Addressing #38  

## Choices

* using the steamapi to get the current build and compare it to the buildId in the file palworld\steamapps\appmanifest_2394010.acf. I am not sure it that is the best way to check. Maybe you have input on this.
* to "update" the server, the script will broadcast a message to the players and restart the server in x minutes. This will only work if the env UPDATE_ON_BOOT is set.
* I have set the default cron expression to once per hour and the warning minutes to 30 minutes

## Test instructions

1. Set  `UPDATE_CRON_EXPRESSION` to `* * * *`
2. Set `WARN_MINUTES` to `1`
3. check in the log that the cronjob is running correctly
4. (Optional) login to the server and create a character
5. change the buildId on line 11 in the file `palworld\steamapps\appmanifest_2394010.acf`` to any other value (for example 13289561)
6. Verify that the server restarts in one Minute after the cronjob has started

**A Real world test is still needed**


Dockerfile
```diff
@@ -42,11 +42,15 @@ ENV PORT= \
     TZ=UTC \
     SERVER_DESCRIPTION= \
     BACKUP_ENABLED=true \
-    BACKUP_CRON_EXPRESSION="0 0 * * *"
+    BACKUP_CRON_EXPRESSION="0 0 * * *" \
+    UPDATE_ENABLED=false \
+    UPDATE_CRON_EXPRESSION="0 * * * *" \
+    WARN_MINUTES=30
```
Dashboy1998, line 48: Since we have a variable called `UPDATE_ON_BOOT` I think these should have the prefix `AUTO_UPDATE_`

scripts/start.sh
```diff
@@ -292,7 +292,15 @@ fi
...
-    echo "$BACKUP_CRON_EXPRESSION bash /usr/local/bin/backup" > "/home/steam/server/crontab"
+    echo "$BACKUP_CRON_EXPRESSION bash /usr/local/bin/backup" >> "/home/steam/server/crontab"
+fi
+
+if [ "${UPDATE_ENABLED}" = true ]; then
+    echo "UPDATE_ENABLED=${UPDATE_ENABLED}"
+    echo "$UPDATE_CRON_EXPRESSION bash /usr/local/bin/update" >> "/home/steam/server/crontab"
+fi
+
+if [ "${UPDATE_ENABLED}" = true ] || [ "${BACKUP_ENABLED}" = true ]; then
```
Dashboy1998, line 303: Since `UPDATE_ON_BOOT` I would also check that before making the cronjob.
```suggestion
if [ "${UPDATE_ENABLED}" = true ] && [ "${UPDATE_ON_BOOT}" = "true" ]; then
    echo "UPDATE_ENABLED=${UPDATE_ENABLED}"
    echo "$UPDATE_CRON_EXPRESSION bash /usr/local/bin/update" >> "/home/steam/server/crontab"
fi

if ([ "${UPDATE_ENABLED}" = true ] && [ "${UPDATE_ON_BOOT}" = "true" ]) || "${BACKUP_ENABLED}" = true ]; then
```

README.md
```diff
@@ -260,6 +263,24 @@ BACKUP_CRON_EXPRESSION is a cron expression, in a Cron-Expression you define an
 Set BACKUP_CRON_EXPRESSION to change the default schedule.
 Example Usage: If BACKUP_CRON_EXPRESSION to `0 2 * * *`, the backup script will run every day at 2:00 AM.
 
+## Configuring Automatic Updates with Cron
+
+To be able to use automatic Updates with this Server the following environment variables **have** to be set to `true`:
+* RCON_ENABLED
+* UPDATE_ON_BOOT
```
Dashboy1998, line 270: I would make it clear this only works if the container restarts.
You are stopping the game server so the container stops and so long as restart is enabled then it automatically updates otherwise you just stop the container.

It is possible to change it so the container says live and the game is updated. However that would require a pretty big change to either `init.sh` or `start.sh`.

scripts/update.sh
```diff
@@ -0,0 +1,24 @@
...
+CURRENTBUILD=$(awk '/buildid/{ print $2 }' < palworld/steamapps/appmanifest_2394010.acf)
+TARGETBUILD=$(curl https://api.steamcmd.net/v1/info/2394010 --silent | grep -P '"public": {"buildid": "\d+"' -o | sed -r 's/.*("[0-9]+")$/\1/')
+
+if [ "$CURRENTBUILD" != "$TARGETBUILD" ]; then
+    echo "New Build was found. Updating the server from $CURRENTBUILD to $TARGETBUILD."
+    if [ "${RCON_ENABLED}" = true ]; then
+        rm /palworld/steamapps/appmanifest_2394010.acf
+        rcon-cli -c /home/steam/server/rcon.yaml "broadcast The_Server_will_update_in_${WARN_MINUTES}_Minutes"
+        sleep "${WARN_MINUTES}m"
+        rcon-cli -c /home/steam/server/rcon.yaml save
```
Dashboy1998, line 17: Probably a good idea to run a backup before the update.
```suggestion
        backup
```

Dashboy1998 (changes requested):

README.md
Dashboy1998, line 270: Should make the linter happy.
```suggestion

* RCON_ENABLED
* UPDATE_ON_BOOT
```

scripts/start.sh
```diff
@@ -292,7 +292,15 @@ fi
 if [ "${BACKUP_ENABLED}" = true ]; then
     echo "BACKUP_ENABLED=${BACKUP_ENABLED}"
     
-    echo "$BACKUP_CRON_EXPRESSION bash /usr/local/bin/backup" > "/home/steam/server/crontab"
+    echo "$BACKUP_CRON_EXPRESSION bash /usr/local/bin/backup" >> "/home/steam/server/crontab"
```
Dashboy1998, line 295: Need `rm -f  "/home/steam/server/crontab"` before this line and outside the if statement.
Without it everytime the container restarts it adds to the file
```
0 0 * * * bash /usr/local/bin/backup
* * * * * bash /usr/local/bin/update
0 0 * * * bash /usr/local/bin/backup
* * * * * bash /usr/local/bin/update
```

scripts/update.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+
+if [ "${UPDATE_ON_BOOT}" = false ]; then
+    echo "Update on Boot needs to be enabled for auto updating"
+    exit 1
+fi
+
+CURRENTBUILD=$(awk '/buildid/{ print $2 }' < palworld/steamapps/appmanifest_2394010.acf)
```
Dashboy1998, line 8: Using relative instead of absolute path.
```
palworld/steamapps/appmanifest_2394010.acf: No such file or directory
```

```suggestion
CURRENTBUILD=$(awk '/buildid/{ print $2 }' < /palworld/steamapps/appmanifest_2394010.acf)
```

Dockerfile
Luatan: Added the prefix `AUTO_UPDATE_`

README.md
Luatan: You're right I totally missed that. Thanks

scripts/update.sh
```diff
@@ -0,0 +1,24 @@
...
+
+if [ "${UPDATE_ON_BOOT}" = false ]; then
+    echo "Update on Boot needs to be enabled for auto updating"
+    exit 1
+fi
+
+CURRENTBUILD=$(awk '/buildid/{ print $2 }' < /palworld/steamapps/appmanifest_2394010.acf)
+TARGETBUILD=$(curl https://api.steamcmd.net/v1/info/2394010 --silent | grep -P '"public": {"buildid": "\d+"' -o | sed -r 's/.*("[0-9]+")$/\1/')
+
+if [ "$CURRENTBUILD" != "$TARGETBUILD" ]; then
```
Dashboy1998, line 11: In the event that `TARGET_BUILD` is empty you shouldn't update.
This will happen if the site is unavailable. You can test by changing to an invalid site.
```suggestion
if [ -n "$TARGETBUILD" ] && [ "$CURRENTBUILD" != "$TARGETBUILD" ]; then
```

README.md
```diff
@@ -177,6 +177,9 @@ It is highly recommended you set the following environment values before startin
 | BACKUP_ENABLED | Enables automatic backups | true | true/false |
 | DELETE_OLD_BACKUPS | Delete backups after a certain number of days                                                                                                                                                       | false          | true/false                                                                                                 |
 | OLD_BACKUP_DAYS    | How many days to keep backups                                                                                                                                                                       | 30             | any positive integer                                                                                       |
+| UPDATE_CRON_EXPRESSION  | Setting affects frequency of automatic updates. | 0 0 \* \* \* | Needs a Cron-Expression - See [Configuring Automatic Backups with Cron](#configuring-automatic-backups-with-cron) |
+| UPDATE_ENABLED | Enables automatic updates | true | true/false |
+| WARN_MINUTES | How long to wait to update the server, after the player were informed. | 30 | !0 |
```
Dashboy1998, line 182: Need to update variable names here too.

scripts/start.sh
Luatan: It should work now as expected. Deleting the crontab before Backup and Update If statement

scripts/update.sh
Luatan: Yes that's a good point. Didn't think of this case.

README.md
```diff
@@ -260,6 +263,27 @@ BACKUP_CRON_EXPRESSION is a cron expression, in a Cron-Expression you define an
...
 Example Usage: If BACKUP_CRON_EXPRESSION to `0 2 * * *`, the backup script will run every day at 2:00 AM.
 
+## Configuring Automatic Updates with Cron
+
+To be able to use automatic Updates with this Server the following environment variables **have** to be set to `true`:
+
+* RCON_ENABLED
+* UPDATE_ON_BOOT
+
+you will also need to use the restart policy `always` or `unless-stopped`. The example docker run command and docker compose file in [How to Use](#how-to-use) already use the needed policy
```
Dashboy1998, line 273:
```suggestion
> [!IMPORTANT]
> 
> If docker restart is not set to policy `always` or `unless-stopped` then the server will shutdown and will need to be
> manually restarted. 
> 
> The example docker run command and docker compose file in [How to Use](#how-to-use) already use the needed policy
```

README.md
Luatan: Updated the readme

README.md
```diff
@@ -260,6 +263,32 @@ BACKUP_CRON_EXPRESSION is a cron expression, in a Cron-Expression you define an
...
+
+* RCON_ENABLED
+* UPDATE_ON_BOOT
+
+> [!IMPORTANT]
+> 
+> If docker restart is not set to policy `always` or `unless-stopped` then the server will shutdown and will need to be
+> manually restarted. 
+> 
+> The example docker run command and docker compose file in [How to Use](#how-to-use) already use the needed policy
```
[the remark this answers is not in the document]
Luatan: Np

README.md
```diff
@@ -177,6 +177,9 @@ It is highly recommended you set the following environment values before startin
 | BACKUP_ENABLED | Enables automatic backups | true | true/false |
 | DELETE_OLD_BACKUPS | Delete backups after a certain number of days                                                                                                                                                       | false          | true/false                                                                                                 |
 | OLD_BACKUP_DAYS    | How many days to keep backups                                                                                                                                                                       | 30             | any positive integer                                                                                       |
+| AUTO_UPDATE_CRON_EXPRESSION  | Setting affects frequency of automatic updates. | 0 0 \* \* \* | Needs a Cron-Expression - See [Configuring Automatic Backups with Cron](#configuring-automatic-backups-with-cron) |
+| AUTO_UPDATE_ENABLED | Enables automatic updates | true | true/false |
```
Dashboy1998, line 181: `AUTO_UPDATE_ENABLED` is set to false in the dockerfile.
```suggestion
| AUTO_UPDATE_ENABLED | Enables automatic updates | false | true/false |
```

Luatan: Yes first wanted to make it default true but since it's not yet tested enough false is better. Forgot to change that.

README.md
```diff
@@ -177,6 +177,9 @@ It is highly recommended you set the following environment values before startin
 | BACKUP_ENABLED | Enables automatic backups | true | true/false |
 | DELETE_OLD_BACKUPS | Delete backups after a certain number of days                                                                                                                                                       | false          | true/false                                                                                                 |
 | OLD_BACKUP_DAYS    | How many days to keep backups                                                                                                                                                                       | 30             | any positive integer                                                                                       |
+| AUTO_UPDATE_CRON_EXPRESSION  | Setting affects frequency of automatic updates. | 0 0 \* \* \* | Needs a Cron-Expression - See [Configuring Automatic Backups with Cron](#configuring-automatic-backups-with-cron) |
```
Dashboy1998, line 180: Should the default value be daily or hourly? Dockerfile is hourly.

scripts/update.sh
```diff
@@ -0,0 +1,24 @@
...
+        rm /palworld/steamapps/appmanifest_2394010.acf
+        rcon-cli -c /home/steam/server/rcon.yaml "broadcast The_Server_will_update_in_${AUTO_UPDATE_WARN_MINUTES}_Minutes"
+        sleep "${AUTO_UPDATE_WARN_MINUTES}m"
+        backup
+        rcon-cli -c /home/steam/server/rcon.yaml "shutdown 1"
+    else
+        echo "An update is available however auto updating without rcon is not supported"
+    fi
+else
+    echo "The Server is up to date!"
```
Dashboy1998, line 23: Not sure if it matters but when the site is not reachable we still say `The Server is update to date`

Dashboy1998 (approved):
Works for me.

README.md
Luatan: I changed it to hourly in the docs

scripts/update.sh
Luatan: I changed the script. Now it also checks what the status code of the curl command is. So we can give better error messages and exit the script if there is no BuildID or the server response does not have the status code 200.

Luatan:
> Works for me.

thank you for the review and fixing by bad code. Great Input. You're a legend 💯

Dashboy1998:
> > Works for me.
> 
> thank you for the review and fixing my bad code. Great Input, you're a legend 💯

You're code was pretty good already. I just pointed out a few typos and mentioned a few edge cases.

Dashboy1998 (approved):

thijsvanloef:
@Luatan thanks for your work!

thijsvanloef:
@Luatan Tested it locally ran into no issues,  thank you for your amazing work, merging :)

PR #216, opened 2024-01-30
Removed root processes

Dashboy1998:
## Context 

* Continuation of #167 
* Allows container to be ran with `--user` but still runs with PUID & GUID.

## Choices

* Keeping the ability to still run as root as to not break anything. Ideally we would later change the image so PUID/PGID is not used and --user must be used.

## Test instructions

> [!IMPORTANT]
> You must set user to `NUMBERICAL_UID:NUMBERICAL_GID`
> To find your UID run `id -u` & to find your GID run `id -g`. In the examples they are listed as 1000:1000.
> If you wish to run it as a different UID/GID this can by done by changing the ownership: `chown UID:GID palworld/`
> or by changing the permissions for all other: `chmod o=rwx palworld/`

1. Try to start the container as root:root, It should work
2. Try to start the container as root:steam, Given error
3. Try to start the container as steam:root, Given error
4. Start container with 1000:1000 (This is the same as steam:steam)
5. Verify rcon-cli is working: `docker exec palworld-server rcon-cli showplayers`
6. Verify backup works: `docker exec palworld-server backup`
7. Verify no processes are running as root: `docker exec -t palworld-server top -n 1`
8. Verify supercronic is running: `docker exec -t palworld-server bash -c "top -n 1 | grep --color supercronic"`


Dashboy1998:
Still in progress.

thijsvanloef:
Would it be possible to port these changes to a true rootless container? Like the one https://github.com/CM2Walki/steamcmd offers?

Dashboy1998:
> Would it be possible to port these changes to a true rootless container? Like the one https://github.com/CM2Walki/steamcmd offers?

Yes.

Now if we do that we have two options.
1. Only start as steam (1000:1000) however this means the files that are created are owned by 1000:1000 in your binds so you must make sure 1000 has permission in the bind.
2. Allow them to start as any nonroot user. Just need to remove the su and all root specific tasks.

Dashboy1998:
@thijsvanloef There is no advantage to switching from cm2network/steamcmd:root to cm2network/steamcmd:steam as we would have to change to root in the dockerfile to install packages then switch back to steam.

If you want to move forward on dropping PUID/PGUID and running with `user: 0:0` then let me know and I can implement that.

thijsvanloef:
>  There is no advantage to switching from cm2network/steamcmd:root to cm2network/steamcmd:steam as we would have to change to root in the dockerfile to install packages then switch back to steam.

Yeah I believe that is why I opened this repo with the base image with the `:root` tag.

I want to keep the PUID/PGUID as that keeps it backwards compatible, and close to the linuxserver.io standard. 
I would much rather keep the PUID PGID and drop the user: 0:0 from the examples as that confuses people

thijsvanloef (approved):

PR #218, opened 2024-01-30
Add initial artifacthub-repo.yml

Twinki14:
# Motivations
Our helm chart should be good to go to put on artifact hub, https://artifacthub.io/

# Modifications
- Add `artifacthub-repo.yml` to `gh-pages`, which registers our helm chart to be picked up by artifact-hub

### DIFF
--- a/artifacthub-repo.yml
+++ b/artifacthub-repo.yml
@@ -1,3 +1,4 @@
-repositoryID: '' # Blank for now
+repositoryID: 32c45b50-f5e4-4e14-a0db-96213c6657ab
 owners:
 - name: thijsvanloef
+- name: Twinki14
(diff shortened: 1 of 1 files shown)


thijsvanloef:
@Twinki14 where does the `repositoryID: 32c45b50-f5e4-4e14-a0db-96213c6657ab` come from?

Twinki14:
> @Twinki14 where does the `repositoryID: 32c45b50-f5e4-4e14-a0db-96213c6657ab` come from?

When adding a repository to artifact hub, you create a repository at https://artifacthub.io, which gives you a repositoryID to use. As far as I know, this repository doesn't give you any "control" over the actual package (aside from the display name of the artifact), it's only there to verify that the publisher is from the same source as the `artifacthub-repo.yml`

Issue #210, opened 2024-01-29
Cannot put URL in SERVER_NAME or SERVER_DESCRIPTION in 0.19.0


bostrt:
## Describe the bug

Server name and description are not set when they contain forward slash `/` in 0.19.0.

## To Reproduce

Configure `SERVER_NAME` or `SERVER_DESCRIPTION` to contain a forward slash `/`, for example a URL and then try to launch container. The name and description will still be default values after starting. 

## Expected behavior

I'd expect to be able to put any sort of character in the server name and description. 

## Logs

Sed error during startup:

```
*****CHECKING FOR EXISTING CONFIG*****
SERVER_NAME="http://palworld.lol 3 | OPEN 24/7 Dedicated"
sed: -e expression #1, char 40: unknown option to `s'
SERVER_DESCRIPTION="Server open to all! More info at http://palworld.lol or https://discord.gg/FYwXY5Ca"
sed: -e expression #1, char 87: unknown option to `s'
```

## docker-compose.yml contents

```yaml
services:
   palworld:
      image: thijsvanloef/palworld-server-docker:latest
      pull_policy: always
      restart: unless-stopped
      container_name: palworld-server
      ports:
        - 8213:8213/udp
        - 27017:27017/udp
      environment:
         - PUID=1000
         - PGID=1000
         - PORT=8213
         - QUERY_PORT=27017
         - PLAYERS=8
         - MULTITHREADING=true
         - RCON_ENABLED=true
         - RCON_PORT=25575
         - ADMIN_PASSWORD="xxxxxxxxxxxxxx"
         - COMMUNITY=true
         - SERVER_NAME="http://palworld.lol 3 | OPEN 24/7 Dedicated"
         - REGION=NA
         - SERVER_DESCRIPTION="Server open to all! More info at http://palworld.lol or https://discord.gg/FYwXY5Ca"
      volumes:
         - ./palworld:/palworld/:z

```

## Additional context

`sed` is interpreting the forward slash inside the server name or description as a separator (e.g. the end of the replacement value) and erroring out.

Here's when I run it manually inside the container:
```
# if [ -n "${SERVER_NAME}" ]; then     echo "SERVER_NAME=${SERVER_NAME}";     sed -E -i "s/ServerName=\"[^\"]*\"/ServerName=$SERVER_NAME/" /palworld/Pal/Saved/Config/LinuxServer/PalWorldSettings.ini; fi

[...]

PR #222, opened 2024-01-30
Fix incorrect environment variable handling

beomseok-park:
## Context 

* Fix incorrect environment variable handling

## Choices

* In `PalWorldSettings.ini`, `ServerName`, `ServerDescription`, `AdminPassword`, `ServerPassword`, `PublicIP`, `Region`, and `BanListURL` should be enclosed in double quotes.
* Fix the incorrect handling of double quotes in `start.sh`.

## Test instructions

1. Stop the Palworld server.
2. Build the container.
3. Delete `<mount_folder>/Pal/Saved/Config/LinuxServer/PalWorldSettings.ini`.
4. Set a `SERVER_NAME` with spaces in the .env. e.g. SERVER_NAME="My Palworld Server"
5. Start the Palworld server.
6. Check the server name in the game.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,15 +57,15 @@ services:
          - PGID=1000
          - PORT=8211 # Optional but recommended
          - PLAYERS=16 # Optional but recommended
-         - SERVER_PASSWORD="worldofpals" # Optional but recommended
+         - SERVER_PASSWORD=worldofpals # Optional but recommended
          - MULTITHREADING=true
          - RCON_ENABLED=true
          - RCON_PORT=25575
          - TZ=UTC
-         - ADMIN_PASSWORD="adminPasswordHere"
+         - ADMIN_PASSWORD=adminPasswordHere
          - COMMUNITY=false  # Enable this if you want your server to show up in the community servers tab, USE WITH SERVER_PASSWORD!
-         - SERVER_NAME="World of Pals"
-         - SERVER_DESCRIPTION="Awesome World of Pal"
+         - SERVER_NAME=World of Pals
+         - SERVER_DESCRIPTION=palworld-server-docker by Thijs van Loef
       volumes:
          - ./palworld:/palworld/
 ```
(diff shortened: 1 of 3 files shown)


thijsvanloef (changes requested):
If the strings are already quoted, like in the examples:
```yaml
         - ADMIN_PASSWORD="adminPasswordHere"
         - SERVER_NAME="World of Pals"
         - SERVER_DESCRIPTION=""
 ```
 Then the whole PalWorldSettings.ini gets really wierd with quotes:
 ```ini
 ServerName=""World of Pals"",ServerDescription="""",AdminPassword=""adminPasswordHere"",ServerPassword=""worldofpals"",PublicPort=8211,PublicIP="",RCONEnabled=true,RCONPort=25575,Region=""
 ```
 A check needs to be implemented to see if the vars already have quotes, and if they do not already have quotes then they should be added.

**Comment by llnut** [comment]
Change the way environment variables to use the following syntax can fix it:
```
environment:
  ADMIN_PASSWORD: "adminPasswordHere"
  SERVER_NAME: "World of Pals"
  SERVER_DESCRIPTION: ""
```
Although this may not be good enough, alternatively, can use sed for additional processing [like this](https://github.com/thijsvanloef/palworld-server-docker/issues/210).

beomseok-park (commented):
I have pushed a new commit that resolves the issue.

Dashboy1998:
> If the strings are already quoted, like in the examples:
> 
> ```yaml
>          - ADMIN_PASSWORD="adminPasswordHere"
>          - SERVER_NAME="World of Pals"
>          - SERVER_DESCRIPTION=""
> ```
> 
> Then the whole PalWorldSettings.ini gets really wierd with quotes:
> 
> ```ini
> ServerName=""World of Pals"",ServerDescription="""",AdminPassword=""adminPasswordHere"",ServerPassword=""worldofpals"",PublicPort=8211,PublicIP="",RCONEnabled=true,RCONPort=25575,Region=""
> ```
> 
> A check needs to be implemented to see if the vars already have quotes, and if they do not already have quotes then they should be added.

[Docker Documentation](https://docs.docker.com/compose/compose-file/05-services/#environment) doesn't support `var="value"` but rather:
* `var: "value"`
* `var: value`
* `var=value`
It is worth noting that you can do `"var=value"` but it's not mentioned

thijsvanloef (approved):
thank you for your work. I've only updated the examples.

JustusPan:
does this mean we should change our docker compose file manually to remove all double quote of values when update to 0.19.1?

thijsvanloef:
@JustusPan Yes, but a check has been put in place that it should work either way

PR #235, opened 2024-01-31
Use storageClassName for pvc

win5923:
## Context 
In this pull request, I aim to address an issue related to the storageClassName in the resources section.

## Test instructions

1. Verify that the storageClassName is correctly set based on the provided size.
2. Ensure that when storageClassName is "-", it is properly handled and results in an empty storageClassName.

### DIFF
--- a/charts/palworld/templates/pvcs.yaml
+++ b/charts/palworld/templates/pvcs.yaml
@@ -27,5 +27,10 @@ spec:
   resources:
     requests:
       storage: {{ .Values.server.storage.size }}
+{{- if .Values.server.storage.storageClassName }}
+{{- if (eq "-" .Values.server.storage.storageClassName) }}
+  storageClassName: ""
+{{- else }}
   storageClassName: {{ .Values.server.storage.storageClassName }}
-{{ end }}
+{{- end }}
+{{- end }}
--- a/charts/palworld/values.yaml
+++ b/charts/palworld/values.yaml
@@ -26,7 +26,14 @@ server:
     # -- The size of the pvc storage.
     size: 12Gi
     # -- The storage class name.
-    storageClassName: ""
+    ## data Persistent Volume Storage Class
+    ## If defined, storageClassName: <storageClassName>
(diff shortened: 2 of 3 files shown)


charts/palworld/values.yaml
```diff
@@ -26,7 +26,14 @@ server:
...
     # -- The storage class name.
-    storageClassName: ""
+    ## data Persistent Volume Storage Class
+    ## If defined, storageClassName: <storageClassName>
+    ## If set to "-", storageClassName: "", which disables dynamic provisioning
+    ## If undefined (the default) or set to null, no storageClassName spec is
+    ##   set, choosing the default provisioner.  (gp2 on AWS, standard on
+    ##   GKE, AWS & OpenStack)
+    ##
+    # storageClassName: "-"
```
MSpreckels, line 36: Explain this to me: why would you ever want to explicilty set the `storageClassName` to an empty string?

AFAIK If no name is explicitly set then kubernetes will infer the default storage class. Wouldn't make much sense to explicitly set it to an empty string.

charts/palworld/templates/pvcs.yaml
```diff
@@ -27,5 +27,10 @@ spec:
   resources:
     requests:
       storage: {{ .Values.server.storage.size }}
+{{- if .Values.server.storage.storageClassName }}
+{{- if (eq "-" .Values.server.storage.storageClassName) }}
+  storageClassName: ""
```
MSpreckels, line 32: see my other comment

MSpreckels (changes requested):

MSpreckels:
I've revoked my change requests.

I don't understand what you wanna achieve with this PR. IMO, you just made it more complicated to set an empty storage class name.

win5923:
> I've revoked my change requests.
> 
> I don't understand what you want to achieve with this PR. IMO, you just made it more complicated to set an empty storage class name.

I see. Most Helm charts are written this way. If it's unnecessary, then it can be omitted. Thank you.

PR #236, opened 2024-01-31
Update serviceType to ClusterIP

win5923:
In a self-built cluster that does not support the Load Balancer feature, setting it to ClusterIP can prevent it from indefinitely getting stuck in 'pending' during creation.


MSpreckels (approved):
LGTM

PR #238, opened 2024-01-31
Fix incorrect link in zh-CN and kr README.md

win5923:
Fix incorrect link and add zh_CN translations for kubernetes setup.


k8s/readme.md
thijsvanloef, line 1: Not a fan of duplicating the content to translate a single sentence

thijsvanloef (changes requested):

k8s/readme.md
win5923: You mean to split it into two commits?

thijsvanloef: only remove the zh-CN/kr from the k8s/readme.md file, after that, ready to merge

win5923: Done!

thijsvanloef (approved):

PR #244, opened 2024-01-31
Add restore command

seoseonyu:
## Context 

* Add the `restore` command to restore a backup file created using the `backup` command.
* Add the usage of the resotre command to the `README.md` (English and Korean have been written, but Chinese needs additional work)

## Choices

* There is a backup command, but not a restore command, so this is for users who are not familiar with CLI commands.

## Test instructions

1. set the path to the actual save file locally and tested it.
2. added the command to the `docker file` and tested whether it runs normally after building.
3. For safe recovery, we shut down the server, temporarily saved the save file, and created a recovery process in case of an error.

We've done our own error handling and testing for all cases, but if additional testing is needed, we can do it.


seoseonyu:
We'll work on a fix for that error.

seoseonyu:
Resolved all Shell - Lint warnings except for functions for error handling.

Dashboy1998 (changes requested):
You seem to be mixing 2x and 4x spacing for indentation. I would recommend using 4x spacing indentation as init.sh, start.sh, dockerfile, and update.sh, use 4x.

Backup.sh is the only one that uses 8x and docker-compose.yml is the only one that uses 3x. Don't think anyone wants to be git blamed to fix docker-compose.yml

scripts/restore.sh
```diff
@@ -0,0 +1,114 @@
...
+    trap 'term_error_handler' ERR
+
+    printf "\e[0;32m*****SHUTDOWN SERVER*****\e[0m\n"
+    if [ "${RCON_ENABLED}" = true ]; then
+        rcon-cli -c /home/steam/server/rcon.yaml save
+        rcon-cli -c /home/steam/server/rcon.yaml "shutdown 1"
+    else
+        echo "RCON is not enabled. Please enable RCON to use this feature."
+    fi
+    printf "\e[0;32mShutdown complete.\e[0m\n"
```
Dashboy1998, line 62: You should exit the script if RCON is not enabled.
Maybe move the shutdown print statement so it is not displayed when you RCON is not enabled.
```suggestion
    if [ "${RCON_ENABLED}" = true ]; then
        printf "\e[0;32m*****SHUTDOWN SERVER*****\e[0m\n"
        rcon-cli -c /home/steam/server/rcon.yaml save
        rcon-cli -c /home/steam/server/rcon.yaml "shutdown 1"
    else
        echo "RCON is not enabled. Please enable RCON to use this feature. Unable to restore backup."
        exit 1
    fi
    printf "\e[0;32mShutdown complete.\e[0m\n"
```

seoseonyu:
Thanks for the suggestion.

We'll take your suggestion into consideration and make changes.

scripts/restore.sh
```diff
@@ -0,0 +1,114 @@
+#!/bin/bash
+
+# Backup file directory path
+BACKUP_DIRECTORY_PATH="/palworld/backups"
+
+# Resotre path
+RESTORE_PATH="/palworld/Pal"
+
+term_error_handler() {
```
Dashboy1998, line 9: Since it's only called by trap you'll want the disable statement
```suggestion

# shellcheck disable=SC2317
term_error_handler() {
```

scripts/restore.sh
```diff
@@ -0,0 +1,114 @@
...
+    echo "An error occurred during server shutdown."
+    exit 1
+}
+
+restore_error_handler() {
```
Dashboy1998, line 14: Since it's only called by trap you'll want the disable statement
```suggestion

# shellcheck disable=SC2317
restore_error_handler() {
```

Dashboy1998:
The container stops when `su steam -c ./start.sh` stops. `su steam -c ./start.sh` stops when the `PalServer.sh` stops.

I believe running the restore command leads to a race since the restore must finish during limited window in which the server has shutdown but `su steam -c ./start.sh` has not.

seoseonyu:
> The container stops when `su steam -c ./start.sh` stops. `su steam -c ./start.sh` stops when the `PalServer.sh` stops.
> 
> I believe running the restore command leads to a race since the restore must finish during limited window in which the server has shutdown but `su steam -c ./start.sh` has not.

Hmm... When I ran several tests, the restore went fine.

```bash
rcon-cli -c /home/steam/server/rcon.yaml save
rcon-cli -c /home/steam/server/rcon.yaml "shutdown 1"
```

Are you saying that if this command is executed, the recovery should happen within the container's shutdown latency?

The docker in our test environment uses
`--stop-timeout 30`
option was enabled.

i will test this separately by removing the `--stop-timeout` option.

scripts/restore.sh
```diff
@@ -0,0 +1,117 @@
...
+        printf "\e[0;32m*****START RESTORE*****\e[0m\n"
+
+        # Recheck the backup file
+        if [ -f "$BACKUP_FILE" ]; then
+            # Copy the save file before restore
+            if [ -d "$RESTORE_PATH/Saved" ]; then
+                echo "Saves the current state before the restore proceeds."
+                rm -rf "./tmp_save"
+                mkdir -p "./tmp_save"
+                \cp -rf "$RESTORE_PATH/Saved" "./tmp_save/Saved"
```
Dashboy1998, line 80: Instead of making a two fixed temporary directory I'd recommend using mktemp to make a parent for them in backups otherwise it's inside wherever you called the script from: `tmp_parent=$(mkdir -d -p /palworld/backup)`. Or do `tmp_parent="/palworld/backup/restore-"$(date +"%Y-%m-%d_%H-%M-%S")`.

Dashboy1998:
> > The container stops when `su steam -c ./start.sh` stops. `su steam -c ./start.sh` stops when the `PalServer.sh` stops.
> > I believe running the restore command leads to a race since the restore must finish during limited window in which the server has shutdown but `su steam -c ./start.sh` has not.
> 
> Hmm... When I ran several tests, the restore went fine.
> 
> ```shell
> rcon-cli -c /home/steam/server/rcon.yaml save
> rcon-cli -c /home/steam/server/rcon.yaml "shutdown 1"
> ```
> 
> Are you saying that if this command is executed, the recovery should happen within the container's shutdown latency?
> 
> The docker in our test environment uses `--stop-timeout 30` option was enabled.
> 
> i will test this separately by removing the `--stop-timeout` option.

Yes it must happen within the shutdown period however there's no way to know the size of the shutdown period. `--stop-timeout 30` is the limit.

There is a non-zero amount of time between `rcon-cli -c /home/steam/server/rcon.yaml "shutdown 1"` and when the container stops. In that amount of time you must complete the restore. 

Also looking at the script again it doesn't look like it actual waits for the server to stop. rcon-cli exits almost instantly. To prevent this you'd want to add this after the rcon `wait $(pidof PalServer-Linux-Test)`.

seoseonyu:
> > > The container stops when `su steam -c ./start.sh` stops. `su steam -c ./start.sh` stops when the `PalServer.sh` stops.
> > > I believe running the restore command leads to a race since the restore must finish during limited window in which the server has shutdown but `su steam -c ./start.sh` has not.
> > 
> > 
> > Hmm... When I ran several tests, the restore went fine.
> > ```shell
> > rcon-cli -c /home/steam/server/rcon.yaml save
> > rcon-cli -c /home/steam/server/rcon.yaml "shutdown 1"
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Are you saying that if this command is executed, the recovery should happen within the container's shutdown latency?
> > The docker in our test environment uses `--stop-timeout 30` option was enabled.
> > i will test this separately by removing the `--stop-timeout` option.
> 
> Yes it must happen within the shutdown period however there's no way to know the size of the shutdown period. `--stop-timeout 30` is the limit.
> 
> There is a non-zero amount of time between `rcon-cli -c /home/steam/server/rcon.yaml "shutdown 1"` and when the container stops. In that amount of time you must complete the restore.
> 
> Also looking at the script again it doesn't look like it actual waits for the server to stop. rcon-cli exits almost instantly. To prevent this you'd want to add this after the rcon `wait $(pidof PalServer-Linux-Test)`.

Thanks for the feedback.

I will do some more testing on this and think of a workaround.

Dashboy1998:
> > > > The container stops when `su steam -c ./start.sh` stops. `su steam -c ./start.sh` stops when the `PalServer.sh` stops.
> > > > I believe running the restore command leads to a race since the restore must finish during limited window in which the server has shutdown but `su steam -c ./start.sh` has not.
> > > 
> > > 
> > > Hmm... When I ran several tests, the restore went fine.
> > > ```shell
> > > rcon-cli -c /home/steam/server/rcon.yaml save
> > > rcon-cli -c /home/steam/server/rcon.yaml "shutdown 1"
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Are you saying that if this command is executed, the recovery should happen within the container's shutdown latency?
> > > The docker in our test environment uses `--stop-timeout 30` option was enabled.
> > > i will test this separately by removing the `--stop-timeout` option.
> > 
> > 
> > Yes it must happen within the shutdown period however there's no way to know the size of the shutdown period. `--stop-timeout 30` is the limit.
> > There is a non-zero amount of time between `rcon-cli -c /home/steam/server/rcon.yaml "shutdown 1"` and when the container stops. In that amount of time you must complete the restore.
> > Also looking at the script again it doesn't look like it actual waits for the server to stop. rcon-cli exits almost instantly. To prevent this you'd want to add this after the rcon `wait $(pidof PalServer-Linux-Test)`.
> 
> Thanks for the feedback.
> 
> I will do some more testing on this and think of a workaround.
> 
> I can't fix it right now because I'm at work. 😭

Realizing the issue also exists for backups I made this PR #248.

Also `wait $(pidof PalServer-Linux-Test)` won't work in the update script since PalServer-Linux-Test is not a child of your script. You'd have to do a pgrep and tail.

seoseonyu:
As suggested by @Dashboy1998, I removed the `--stop-timeout` and `--restart unless-stopped` options and forced a wait using the `sleep` command in the middle of the restore process and found that the container was actually terminated before the restore was complete.

As per his suggestion, I added code to `init.sh` to suspend container termination until the restore is complete, referring to PR #248. 

1. Force the `restore` script to be long-lived via the `sleep` command and test if it waits for the container to terminate while the script is in progress.
2. test if the game is restored to the point in the `backup` file as normal if the container is waiting for termination.

We ran the above tests and confirmed that the container waits for termination and the game is restored to the point of view of the `backup` file.

We fixed the creation location of the temporary directory used for restoration to `/palworld/backups`.

scripts/restore.sh
```diff
@@ -0,0 +1,120 @@
...
+        if [[ $RUN_ANSWER == "y" ]] || [[ $RUN_ANSWER == "Y" ]]; then
+            rm -rf "$RESTORE_PATH/Saved"
+            mv "$TMP_SAVE_PATH/Saved" "$RESTORE_PATH"
+            printf "\e[0;32mRecovery complete.\e[0m\n"
+        fi
+    fi
+
+    echo "Clean up the temporary directory."
+    rm -rf "$TMP_PATH"
+    rm -rf "$TMP_SAVE_PATH"
```
Dashboy1998, line 32: Can be one line
```suggestion
    rm -rf "$TMP_PATH" "$TMP_SAVE_PATH"
```

scripts/restore.sh
```diff
@@ -0,0 +1,120 @@
...
+    echo "An error occurred during server shutdown."
+    exit 1
+}
+
+# shellcheck disable=SC2317
+restore_error_handler() {
+    printf "\033[0;31mAn error occurred during restore.\033[0m\n"
+    if [ -d "$TMP_SAVE_PATH/Saved" ]; then
+        read -rp "I have a backup before recovery can proceed. Do you want to recovery it? (y/n): " RUN_ANSWER
+        if [[ $RUN_ANSWER == "y" ]] || [[ $RUN_ANSWER == "Y" ]]; then
```
Dashboy1998, line 23: I'd convert to lowercase for comparison.
```suggestion
        if [[ ${RUN_ANSWER,,} == "y" ]]; then
```

README.md
```diff
@@ -247,6 +247,19 @@ This will create a backup at `/palworld/backups/`
...
+To restore a backup, use the command:
+
+```bash
+docker exec palworld-server restore
+```
+
+The container must be restarted after the recovery progresses.
+
+> [!IMPORTANT]
+> The `RCON_ENABLED` environment variable must be `true` to use this command.
```
Dashboy1998, line 261: Modified readme to be more inline with [Configuring Automatic Updates with Cron](https://github.com/thijsvanloef/palworld-server-docker?tab=readme-ov-file#configuring-automatic-updates-with-cron)
```suggestion
The `RCON_ENABLED` environment variable must be set to `true` to use this command.
> [!IMPORTANT]
> If docker restart is not set to policy `always` or `unless-stopped` then the server will shutdown and will need to be
> manually restarted.
>
> The example docker run command and docker compose file in [How to Use](#how-to-use) already uses the needed policy
```

scripts/restore.sh
```diff
@@ -0,0 +1,120 @@
...
+        
+        printf "\e[0;32m*****START RESTORE*****\e[0m\n"
+
+        # Recheck the backup file
+        if [ -f "$BACKUP_FILE" ]; then
+            # Copy the save file before restore
+            if [ -d "$RESTORE_PATH/Saved" ]; then
+                echo "Saves the current state before the restore proceeds."
+                echo "$TMP_SAVE_PATH"
+                mkdir -p "$TMP_SAVE_PATH"
```
Dashboy1998, line 83: I would add a chmod in so the file is owned by steam not root. 
In the event of a container kill or sigterm on restore it will remain.
While it shouldn't be needed since the container `chown -R steam:steam /palworld /home/steam/` it prevents having a directory owned by root from existing. 

```suggestion
                mkdir -p "$TMP_SAVE_PATH"
                if [ "$(id -u)" -eq 0 ]; then
                    chown steam:steam "$TMP_PATH"
                fi
```

scripts/restore.sh
```diff
@@ -0,0 +1,120 @@
...
+
+                while [ ! -d "$TMP_SAVE_PATH/Saved" ]; do
+                    sleep 1
+                done
+
+                printf "\e[0;32mSave complete.\e[0m\n"
+            fi
+            
+            # Create tmp directory
+            TMP_PATH=$(mktemp -d -p "/palworld/backups")
```
Dashboy1998, line 94: I would add a chmod in so the file is owned by steam not root. 
In the event of a container kill or sigterm on restore it will remain.
While it shouldn't be needed since the container `chown -R steam:steam /palworld /home/steam/` it prevents having a directory owned by root from existing. 

```suggestion
            TMP_PATH=$(mktemp -d -p "/palworld/backups")
            if [ "$(id -u)" -eq 0 ]; then
                chown steam:steam "$TMP_PATH"
            fi
```

scripts/restore.sh
seoseonyu: The `TMP_SAVE_PATH` is used to temporarily save the current state of the restore progress and to restore if an error occurs.

It will be deleted when the restore is successful.

Should I grant ownership to a folder that will be deleted anyway?

scripts/restore.sh
seoseonyu: This folder is likewise a temporary directory for extracting the backup `.tar` file. 

It will be deleted upon completion of the restore process or if an error occurs.

Will we ever need to specify ownership?

scripts/restore.sh
Dashboy1998: I would because if an error is never raised and it doesn't complete then it won't be deleted.

This can happen when the container is killed such as when it as exceeded it's graceful stop period or during a power outage.

seoseonyu:
Fixed for suggestions

README.md
```diff
@@ -247,6 +247,21 @@ This will create a backup at `/palworld/backups/`
 
 The server will run a save before the backup if rcon is enabled.
 
+## Restore a backup
+
+To restore a backup, use the command:
+
+```bash
+docker exec palworld-server restore
```
Dashboy1998, line 255:
```suggestion
docker exec -i palworld-server restore
```
This needs to be interactive

docs/kr/README.md
```diff
@@ -198,11 +198,26 @@ docker exec palworld-server backup
 
 rcon이 활성화된 경우 서버는 백업 전에 저장을 실행합니다.
 
+## 백업 복원 하기
+
+백업을 복원하려면 다음 명령을 사용합니다:
+
+```bash
+docker exec palworld-server restore
```
Dashboy1998, line 206:
```suggestion
docker exec -i palworld-server restore
```

This needs to be interactive

Dashboy1998 (changes requested):

README.md
```diff
@@ -247,6 +247,21 @@ This will create a backup at `/palworld/backups/`
 
 The server will run a save before the backup if rcon is enabled.
 
+## Restore a backup
+
+To restore a backup, use the command:
+
+```bash
+docker exec -i palworld-server restore
```
Dashboy1998, line 255:
```suggestion
docker exec -it palworld-server restore
```

I really thought only `-i` was needed but that resulted in this
```
Backup List:
1) /palworld/backups/blah.tar.gz
2) /palworld/backups/palworld-save-2024-02-02_00-21-18.tar.gz
3) /palworld/backups/subfolder/palworld-save-2024-02-02_00-20-50.tar.gz
#? 3
Selected backup: /palworld/backups/subfolder/palworld-save-2024-02-02_00-20-50.tar.gz
Do you want to continue with the command?
y
```

Changing it to `-it` gave the expected

```
Backup List:
1) /palworld/backups/blah.tar.gz
2) /palworld/backups/palworld-save-2024-02-02_00-21-18.tar.gz
3) /palworld/backups/subfolder/palworld-save-2024-02-02_00-20-50.tar.gz
#? 3
Selected backup: /palworld/backups/subfolder/palworld-save-2024-02-02_00-20-50.tar.gz
Do you want to continue with the command?
When you run it, the server will be stopped and the recovery will proceed. (y/n): y
```

docs/kr/README.md
```diff
@@ -198,11 +198,26 @@ docker exec palworld-server backup
 
 rcon이 활성화된 경우 서버는 백업 전에 저장을 실행합니다.
 
+## 백업 복원 하기
+
+백업을 복원하려면 다음 명령을 사용합니다:
+
+```bash
+docker exec -i palworld-server restore
```
Dashboy1998, line 206:
```suggestion
docker exec -it palworld-server restore
```

I really thought only `-i` was needed but that resulted in this
```
Backup List:
1) /palworld/backups/blah.tar.gz
2) /palworld/backups/palworld-save-2024-02-02_00-21-18.tar.gz
3) /palworld/backups/subfolder/palworld-save-2024-02-02_00-20-50.tar.gz
#? 3
Selected backup: /palworld/backups/subfolder/palworld-save-2024-02-02_00-20-50.tar.gz
Do you want to continue with the command?
y
```

Changing it to `-it` gave the expected

```
Backup List:
1) /palworld/backups/blah.tar.gz
2) /palworld/backups/palworld-save-2024-02-02_00-21-18.tar.gz
3) /palworld/backups/subfolder/palworld-save-2024-02-02_00-20-50.tar.gz
#? 3
Selected backup: /palworld/backups/subfolder/palworld-save-2024-02-02_00-20-50.tar.gz
Do you want to continue with the command?
When you run it, the server will be stopped and the recovery will proceed. (y/n): y
```

Dashboy1998 (changes requested):

README.md
seoseonyu: 조금더 잘확인했어야했는데 제실수입니다.
빠른 정정 감사합니다.

Dashboy1998 (approved):
# Tests
Ran store with: `docker exec -it palworld-server restore`, note the `-it`

Created a large file for testing `fallocate -l 10G palworld/Pal/Saved/deleteme`
Created a backup I used to restore

1. With rcon=true, restored a backup from backups/. Restored as expected
1. With rcon=true, restored a backup from backups/subfolder/. Restored as expected
1. Tried to restore with RCON=false, provided message and exited as expected `RCON is not enabled. Please enable RCON to use this feature.`
1. With rcon=true I deleted rcon.yaml to test `term_error_handler`, it triggered as expected
    ```
    *****STARTING PROCESS*****
    *****SHUTDOWN SERVER*****
    cli: config: parse file: read file: open /home/steam/server/rcon.yaml: no such file or directory
    An error occurred during server shutdown.
    ```
1. Tried to restore an invalid backup to caught `restore_error_handler`, it triggered as expected
    ```
    *****START RESTORE*****
    Saves the current state before the restore proceeds.
    /palworld/backups/restore-2024-02-02_00-32-50
    Save complete.

    gzip: stdin: not in gzip format
    tar: Child returned status 1
    tar: Error is not recoverable: exiting now
    An error occurred during restore.
    I have a backup before recovery can proceed. Do you want to recovery it? (y/n): y
    Recovery complete.
    Clean up the temporary directory.
    ```
    ```
    *****START RESTORE*****
    Saves the current state before the restore proceeds.
    /palworld/backups/restore-2024-02-02_00-34-40
    Save complete.

    gzip: stdin: not in gzip format
    tar: Child returned status 1
    tar: Error is not recoverable: exiting now
    An error occurred during restore.
    I have a backup before recovery can proceed. Do you want to recovery it? (y/n): n
    Clean up the temporary directory.
    ```
1. Issued stop to the container during restore and verified it waits until restore is finished

1. I confirmed the container does wait for the restore process to complete before stopping

1. Other noteworthly results. The server must be running in order to restore. If you stop the server while in the restore process but before the shutdown command is issued then then `term_error_handler` is caught. 

    ```
    *****STARTING PROCESS*****
    *****SHUTDOWN SERVER*****
    cli: execute: auth: rcon: dial tcp 127.0.0.1:25575: connect: connection refused
    An error occurred during server shutdown.
    ```

# Proposed solutions for issues
## Proposed solution for 5
Use `file` and `grep` to verify it is a gzip file. `file "${BACKUP_FILE}" | grep -q 'gzip'`

While this won't fix all the restore errors it will validate that we only attempt to restore gzip files.

Mind you `file` is not installed in the container.

## Possible solution for 8 (Just adds more issues)
Use `pgrep` to see if the server is running before running the rcon command

Having this solution would introduce a new issue where if someone attempts to restore while the server is updating then starts resulting in who knows what.

# Conclusion

While I did propose solutions for 5, I think what we have here is great and works as intended. I don't see a huge need to add an additional validation check for the gzip file as the output given by the error was fairly identifiable.

You have put a lot a work into this. It is amazing. I approve these changes.

seoseonyu:
Thank you very much for your help

thijsvanloef (changes requested):
@seoseonyu I really appreciate the work you've put into this. And thanks @Dashboy1998 for testing it so thoroughly. 

I do have 1 single request, when you start the script, please show a disclaimer with the following information:

```
This script has been designed to help you restore; however, I am not responsible for any data loss. It is recommended that you create a backup beforehand, and in the event of script failure, be prepared to restore it manually.
```

I do not want to be responsible for people losing their data due to unforeseen circumstances.

seoseonyu:
I added this disclaimer to the process of selecting the backup file to restore and confirming user intentions just before running the command.

The output looks like the picture below.

image

thijsvanloef (approved):
@seoseonyu Amazing work!

Dashboy1998:
> I haven't executed this script yet but just wanna ask, will this let me select a backup file or will it only restore the latest backup? 

It let's you select a backup.

Issue #243, opened 2024-01-31
Add lifecycle preStop stop hook and probes to Kubernetes


anthr76:
**Is your feature request related to a problem? Please describe.**

Not necessarily a problem but I can see it being one in certain use-cases.

**Describe the solution you'd like**

Make a script to gracefully shutdown Palworld with Rcon via a script in the container with a prestop hook (https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/) and example of this would be something like:

```yaml
    lifecycle:
      preStop:
        exec:
          command:
          - /pal-graceful-shutdown.sh
```

In the script we can simply execute something like `/Shutdown {Seconds} {MessageText}`

This is similar to https://github.com/thijsvanloef/palworld-server-docker/pull/172 albeit a bit more simple

I think it would also be useful to add probes to Kubernetes so in the case something go unhealthy the pod will automatically restart. I think this may be a bit trickier since I don't think Kubernetes supports raw UDP probes like it does with TCP https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#tcp-probes

win5923:
Kubernetes does not support UDP liveness probes.
I think the approach using `preStop` can perform a backup before terminating the pod.

anthr76:
> Kubernetes does not support UDP liveness probes.

+1 I don't think it does either. I suppose is our best bet is to use a command (exec.command) https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-a-liveness-command that test the liveliness in a container

chkpwd:
```yaml
  liveness: &probes
    enabled: true
    custom: true
    spec:
      exec:
        command:
        - bash
        - -c
        - 'pgrep "PalServer-Linux" > /dev/null || exit 1'
      initialDelaySeconds: 60
      periodSeconds: 10
      timeoutSeconds: 1
      failureThreshold: 3
 ```

You can check for this process instead.

PR #246, opened 2024-01-31
[Kubernetes] backup before a container is terminated

win5923:
Incorporate backup functionality before the termination of a container due to API requests or management events such as liveness/startup probe failure, preemption, resource contention, and others.

## Choices

* Add backup before a container is terminated.
* Addressing https://github.com/thijsvanloef/palworld-server-docker/issues/243#issue-2110148331
* Refer to https://github.com/thijsvanloef/palworld-server-docker/pull/206#issue-2105868641

## Test instructions

1. I haven't tested it yet. I'll conduct the test in a few days.

### DIFF
--- a/charts/palworld/templates/deployments.yaml
+++ b/charts/palworld/templates/deployments.yaml
@@ -40,6 +40,18 @@ spec:
         - name: server
           image: "{{ .Values.server.image.name }}:{{ .Values.server.image.tag }}"
           imagePullPolicy: {{ .Values.server.image.imagePullPolicy }}
+          lifecycle:
+            preStop:
+              exec:
+                command: ["/bin/sh", "-c"]
+                args:
+                  - |
+                    cont=$(kubectl -n {{ .Values.namespace }} get pods -o=jsonpath='{.items[?(@.metadata.labels.app\.kubernetes\.io/name=="{{ .Release.Name }}-server")].metadata.name}')
+                    function exec_rcon_cmd() {
+                      kubectl exec -n {{ .Values.namespace }} -i pod/$cont rcon-cli "$1"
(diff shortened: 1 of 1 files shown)


anthr76:
Considering this would need RBAC (and kubectl in the container) it would be probably be better to just execute a script in the container

win5923:
> Considering this would need RBAC (and kubectl in the container) it would be probably be better to just execute a script in the container

I just realized that I am running bash inside a pod, so there is no need to execute kubectl. Thank you for your advice.

anthr76:
Good point 👍🏼 I think we just address the probe(s) in another PR as that may be tricky. I'll try to think about a solution. This looks good! Should we also add a `/Shutdown` to the rcon commands?

win5923:
After a liveness probe failure, the pod will be restarted. I think it may not be necessary. I might conduct a test using Minikube in the next few days.

thijsvanloef:
@win5923 Just give me a headsup when it is ready to merge :)

win5923:
> @win5923 Just give me a headsup when it is ready to merge :)

Done!

bobbyrward:
This doesn't work.  [exec only takes a command](https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#LifecycleHandler) not a command and args

win5923:
> This doesn't work. [exec only takes a command](https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#LifecycleHandler) not a command and args

Oh, you're right. I forgot to save before applying. No wonder I was able to apply successfully.
I will resubmit a new pull request to correct this error. Thank you very much.

PR #258, opened 2024-02-01
Add quotes to rcon.yaml password

BongoEADGC6:
Quotes allow for special characters to be used in the password field for the rcon config yaml.

## Context 

* Allow or the use of special characters in rcon and admin passwords
## Choices

* This is what yaml allows 
## Test instructions

Tested with and without special characters.
1. Set env vars
2. Run EOL text insertion
3.  run `rcon-cli` to check errors

* [X] I've added documentation about this change to the README. (Documentation shouldn't be required as there is no documentation stating a limitation.)

### DIFF
--- a/scripts/start.sh
+++ b/scripts/start.sh
@@ -320,7 +320,7 @@ fi
 cat >/home/steam/server/rcon.yaml  <<EOL
 default:
   address: "127.0.0.1:${RCON_PORT}"
-  password: ${ADMIN_PASSWORD}
+  password: '${ADMIN_PASSWORD}'
 EOL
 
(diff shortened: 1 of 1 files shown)


Dashboy1998:
What special characters currently don't work?
* ADMIN_PASSWORD=sp!@#$%^*()!1-_=+[]{}`~:;<>,.?SPACE WORD
The above works in main now. Looks like &/\ don't work in either.

thijsvanloef (approved):
While this does not change the actual string in the file, this does resolve an uncaught shellcheck error: https://www.shellcheck.net/wiki/SC2086

Issue #259, opened 2024-02-01
Restart Container to Stop Memory Leaks


mobileqaguy:
**Is your feature request related to a problem? Please describe.**
After playing for several hours, things start functioning incorrectly. Sometimes I cant pickup a pal to set it to do a specific task, acquire items made on assembly lines, and other oddities. I've found that simply restarting the container helps a lot. I see in the latest release that there was a k8s setting for a daily reboot but nothing to do scheduled reboots.

**Describe the solution you'd like**
I'd like a way to be able to set environment variable for this. Ideally, use a cron expression for when to run it. 

**Describe alternatives you've considered**
I have been just restarting it myself

**Additional context**
It seems that restarting the container almost always fixes the weird issues i encounter. Below is the script that I am using.  My cron expression is `15 1,5,9,13,17,21 * * *` since these times fall in line the most with people not being on the server.

```
#!/bin/bash

FIRST_WARNING_MINUTES=5
FINAL_WARNING_MINUTES=1
DOCKER_EXEC_RCON_COMMAND_PREPEND="docker exec -it palworld-server rcon-cli"

sleepForMinutes() {
    local minutes="$1"
    local seconds="$(( minutes * 60 ))"
    
    if [ "$minutes" -gt 1 ]; then
        seconds="$(( seconds - 60 ))"
    fi
    
    sleep "$seconds"
}

getRconCliBroadcastRestartString() {
    local warning_minutes="$1"

    if [ "$warning_minutes" -gt 1 ]; then
        echo "$DOCKER_EXEC_RCON_COMMAND_PREPEND 'Broadcast Server_rebooting_in_${warning_minutes}_minutes'"
    elif [ "$warning_minutes" -eq 1 ]; then
        echo "$DOCKER_EXEC_RCON_COMMAND_PREPEND 'Broadcast Server_rebooting_in_${warning_minutes}_minute'"
    else
        echo "$DOCKER_EXEC_RCON_COMMAND_PREPEND 'Broadcast Server_rebooting_now'"
    fi
}

sendRconCliCommand() {
    local command="$1"
    eval "${command}"
    if [ $? -ne 0 ]; then
        exit 1
    fi
}

restartServer() {
    docker restart palworld-server
}

saveServer() {
    sendRconCliCommand "${DOCKER_EXEC_RCON_COMMAND_PREPEND} Save"
}

saveAndRestart() {

    # Send broadcast message to server letting them know that the server will restart in 5 minutes
[...]

PR #260, opened 2024-02-02
Reboot cron job

Dashboy1998:
## Context 

* Adds a cron job to automatically reboot the server as mentioned in #259 

## Choices

* Already using cron for automatic updates and backups

## Test instructions

1. Enable rcon & auto reboots 
2. Set the reboot expression near your current time so you don't wait long
3. Start the container
4. Join the game and wait for the reboot

image


README.md
```diff
@@ -290,6 +293,30 @@ AUTO_UPDATE_CRON_EXPRESSION is a cron expression, in a Cron-Expression you defin
...
+
+Set AUTO_REBOOT_ENABLED enable or disable automatic backups (Default is disabled)
+
+AUTO_REBOOT_CRON_EXPRESSION is a cron expression, in a Cron-Expression you define an interval for when to run jobs.
+
+> [!TIP]
+> This image uses Supercronic for crons
+> see [supercronic](https://github.com/aptible/supercronic#crontab-format)
+> or
+> [Crontab Generat](https://crontab-generator.org).
```
Technetium1, line 315: @Dashboy1998 Should be `Generator`

Technetium1:
Could be nice to add in https://github.com/thijsvanloef/palworld-server-docker/blob/main/.env.example

MSpreckels:
Dont we already have a reboot cronjob? #206

Dashboy1998:
> Dont we already have a reboot cronjob? #206

That appears to be specifically for Kubernetes whereas this will work for the stand alone container.

MSpreckels:
You are correct. Does it make sense to keep the k8s only variant of a daily reboot when it will be implemented in the container?

Dashboy1998:
> You are correct. Does it make sense to keep the k8s only variant of a daily reboot when it will be implemented in the container?

I think the k8s version has some specific benefits for k8s. I don't use k8s so I'm not really sure of the benefits of having a separate solution for them.

thijsvanloef (approved):
LGTM

Technetium1:
@MSpreckels keeping both makes sense to me because they are not restarting _exactly_ the same way. There's more granularity available with the Kubernetes spec, such as [killing the job if it times out](https://github.com/thijsvanloef/palworld-server-docker/blob/24c4071ba1fa690ffe98011680ea680f1ef247e7/k8s/examples/cronjob-reboot/cronjob.yaml#L11), and the fact that prevents concurrent jobs from executing at once, but the same effect is unnecessary with only 1 Docker container. Even moreso because there's potential for that job to run on multiple replicas in a cluster (even if they don't exist right now); with offset times, or just waiting for completion on each replica.

Worth noting that `AUTO_REBOOT_ENABLED=false` to prevent issue for k8s users.

PR #264, opened 2024-02-02
Updated .env.example

Dashboy1998:
## Context 

* Added missing environmental variables
* @Technetium1 reminded me to add variables to the env file.

## Choices

* .env.example should probably contain all variables and their default values.

## Test instructions

1. Verify container uses the newly added variables from the .env file


Technetium1 (approved):

thijsvanloef:
Removed the quotes around the crontab expressions, to keep it in line with not quoting strings

Issue #32, opened 2024-01-21
Can we have option to stop the server if last player disconnected


nicedevil007:
**Is your feature request related to a problem? Please describe.**
Pets may die during the time where nobody is looking for them (some people have to work :D)
On other steamcmd servers it was possible to set a flag to stop the server after the last player disconnected, I believe it was satisfactory, but don't hurt me if I'm wrong.

**Describe the solution you'd like**
We need an ENV variable that do the feature for us.

**Describe alternatives you've considered**
n.a.

**Additional context**
all said, at least I hope so :D

thijsvanloef:
Would be a great feature!

Twanislas:
I support this, it would also prevent the server to reach huge number in days xD

I do think this might as well be implemented at the game level to have a feature to "freeze time" once the last player disconnects.

fryfrog:
You could use something via RCON to watch the status of `/ShowPlayers` maybe? At least, it could shutdown when 0 players. The user would have to start it back up.

Twanislas:
That's indeed an excellent workaround in the meantime, I might try and have a stab at it (and open a PR!) if I find the free cycles :)

Cheers !

fremus89:
[script.txt](https://github.com/thijsvanloef/palworld-server-docker/files/14057720/script.txt)
Something like this do I have in mind, but is untested

fryfrog:
I'd use `rcon` to list the players too.

fremus89:
```shell
#!/bin/bash

# Function to check players and take action based on the response
check_players() {
    players_output=$(docker exec -it palworld-server rcon-cli ShowPlayers)

    # Check if the response contains the header followed by an empty line
    if echo "$players_output" | grep -q "name,playeruid,steamid"$'\n'$'\n'; then
        echo "No players found. Shutting down the server."
        docker exec -it palworld-server rcon-cli Shutdown
    else
        echo "Players found. Server will continue running."
        # You can add additional actions or commands here if needed
    fi
}

check_players

``` 

this in a nice little loop with a delay et voilà


to start the server i have something similar, unfortunately, the server has to be reached twice, as the game times out during first connection and server boot.
```shell
#!/bin/bash

TARGET_PORT=8211

while true; do
    # Use tcpdump to capture incoming traffic on the target port
    capture_result=$(sudo tcpdump -n -c 1 -i any port $TARGET_PORT 2>/dev/null)

    # Check if any packets were captured
[...]

PR #273, opened 2024-02-02
Pause without players

Luatan:
## Context 

* Adds the ability to pause the server
* Addressing #32 

## Choices

* Added an additional Script that will send a SIGSTOP to the server. After the process is suspended, it will monitor the server port for any connection attempts.
* If a player connects to the port it will resume the process and the player will be able to spawn without a timeout.

**Feedback is appreciated, if you see something that can be improved!**

### Todo

* Add Cronjob
* RCON handling if the server is suspended
* Test how suspending the process reflects on the Ingame time
* Check if port is set

### Manually
1. Start the server with the default docker compose file
2. run `docker exec palworld-server pause`
3. check the server log for a message, that the server has been paused
4. run `echo>/dev/udp/127.0.0.1/8211` or start the game and login
5. check that the server has continued

### Cron
1. todo

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,6 +10,7 @@ LABEL maintainer="thijs@loef.dev" \
 RUN apt-get update && apt-get install -y --no-install-recommends \
     procps=2:3.3.17-5 \
     wget=1.21-1+deb11u1 \
+    tcpdump=4.99.0-2+deb11u1 \
     xdg-user-dirs=0.17-2 \
     && apt-get clean \
     && rm -rf /var/lib/apt/lists/*
--- a/Dockerfile
+++ b/Dockerfile
@@ -72,7 +72,8 @@ COPY ./scripts/* /home/steam/server/
 RUN chmod +x /home/steam/server/*.sh && \
     mv /home/steam/server/backup.sh /usr/local/bin/backup && \
     mv /home/steam/server/update.sh /usr/local/bin/update && \
-    mv /home/steam/server/restore.sh /usr/local/bin/restore
+    mv /home/steam/server/restore.sh /usr/local/bin/restore && \
+    mv /home/steam/server/pause.sh /usr/local/bin/pause
 
 WORKDIR /home/steam/server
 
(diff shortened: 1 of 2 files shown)


Dashboy1998:
As an alternative to using SIGSTOP you could shutdown the server, monitor the port then start the server.

Luatan:
> As an alternative to using SIGSTOP you could shutdown the server, monitor the port then start the server.

I think if I stop the server binary, I also stop the container because start.sh will exit.

Dashboy1998:
> I think if I stop the server binary, I also stop the container because start.sh will exit.

As written yes. With a few modifications to start.sh you could change that behavioral.

Luatan:
> > I think if I stop the server binary, I also stop the container because start.sh will exit.
> 
> As written yes. With a few modifications to start.sh you could change that behavioral.

Do you think it would make sense to make the pal server into a service so we can start/restart/stop it easier? I think this would also help with other scripts for example the auto-updater

Aside from that. I am not sure if a restart would be fast enough that you don't have a timeout.

Dashboy1998:
> Do you think it would make sense to make the pal server into a service so we can start/restart/stop it easier? If that's even possible inside the container.

I'm not sure how running as a service would make a difference.

You're probably right, the time to start the server might be too long. However SIGSTOP would cause issues with autoreboot.sh, backup.sh, and update.sh. Those don't check if the rcon command was successful, they assume it was. 

The term handler in init.sh would produce an error for rcon too and it would hang as the server would never stop since the rcon stop would never be recieved by the paused server.

While it would prevent restore.sh from working properly, that wouldn't have an big issues as @seoseonyu did a great job with error handlers.

That isn't to say this isn't doable. It definitely is doable but it would be a big PR.

I would say something making that many changes should be multiple PRs due to the error handling, rcon checking, and status checking we would need to add to autoreboot.sh, backup.sh, and update.sh.

Luatan:
closing this. I think it is better for an external Script to pause the container instead of trying to stuff this in (at least at the moment). Maybe the game will give us a command in a later version.

Issue #275, opened 2024-02-02
Environmental Var for AdminPassword needs to be in ""


chaoscode:
## Describe the bug

I had the manually edit the ini file to put the "" around the password to get it to work.

ThePrinoob:
Wanted to report that too. I had a & in my password and it always only wrote everything before. 
Using v0.21.0.

win5923:
> Wanted to report that too. I had a & in my password and it always only wrote everything before. Using v0.21.0.

In Shell scripting, `&` is a special character used.
Not recommended to set a variable name as `&` because it may lead to confusion or unexpected behavior due to its special meaning in Shell syntax.

win5923:
> had the manually edit the ini file to put the "" around the password to get it to work.

If you want to become an administrator, you must set the `AdminPassword`. If you simply don't want anyone to become an administrator, I can create a pull request (PR) to allow setting ADMIN_PASSWORD to an empty string.

win5923:
> had the manually edit the ini file to put the "" around the password to get it to work.

If you want to become an administrator, you must set the `AdminPassword`. If you simply don't want anyone to become an administrator, I can create a PR to allow setting `AdminPassword` to an empty string through env `ADMIN_PASSWORD`

thijsvanloef:
Sadly cannot get around the & at the moment, this is a limitation on using bash to set the settings. Hopefully https://github.com/thijsvanloef/palworld-server-docker/pull/207  is able to fix this later on

MSpreckels:
Hey there does this issue persists after the latest update?

chaoscode:
This is fixed. Thanks.

PR #281, opened 2024-02-03
Fix: with no `SERVER_PASSWORD` can still apply setting

win5923:
## Choices
*  Allow setting `ServerPassword` to an empty string through env `SERVER_PASSWORD`
* Resolve the issue of applying settings without manually modifying PalWorldSettings.ini when there is no server password.
* Addressing a similar issue as described in https://github.com/thijsvanloef/palworld-server-docker/issues/275#issue-2115930445.

## Test instructions

1. Comment out the SERVER_PASSWORD or set it to an empty string.
2. Open the game and check if the `ServerPassword` in PalWorldSettings.ini is set to an empty string.

### DIFF
--- a/scripts/start.sh
+++ b/scripts/start.sh
@@ -122,7 +122,7 @@ if [ -n "${SERVER_PASSWORD}" ]; then
     SERVER_PASSWORD=$(sed -e 's/^"\(.*\)"$/\1/' -e "s/^'\(.*\)'$/\1/" <<< "$SERVER_PASSWORD")
     echo "SERVER_PASSWORD=${SERVER_PASSWORD}"
     sed -E -i "s/ServerPassword=\"[^\"]*\"/ServerPassword=\"$SERVER_PASSWORD\"/" /palworld/Pal/Saved/Config/LinuxServer/PalWorldSettings.ini
-elif [ -v SERVER_PASSWORD ]; then
+elif [ -v "${SERVER_PASSWORD}" ]; then
     echo "No SERVER_PASSWORD"
     sed -E -i "s/ServerPassword=\"[^\"]*\"/ServerPassword=\"\"/" /palworld/Pal/Saved/Config/LinuxServer/PalWorldSettings.ini
 fi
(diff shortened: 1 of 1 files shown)


win5923:
Do I need to apply this setting to `ADMIN_PASSWORD` as well?

Dashboy1998:
> Do I need to apply this setting to `ADMIN_PASSWORD` as well?

[rcon-cli](https://github.com/gorcon/rcon-cli/issues/37) doesn't work with an empty password. 
RCON protocol itself refuses all requests when no password is set.

If there is no admin password then you aren't able to become an admin: `/adminpassword ""`

scripts/start.sh
```diff
@@ -122,6 +122,9 @@ if [ -n "${SERVER_PASSWORD}" ]; then
     SERVER_PASSWORD=$(sed -e 's/^"\(.*\)"$/\1/' -e "s/^'\(.*\)'$/\1/" <<< "$SERVER_PASSWORD")
     echo "SERVER_PASSWORD=${SERVER_PASSWORD}"
     sed -E -i "s/ServerPassword=\"[^\"]*\"/ServerPassword=\"$SERVER_PASSWORD\"/" /palworld/Pal/Saved/Config/LinuxServer/PalWorldSettings.ini
+else
+    echo "No SERVER_PASSWORD"
+    sed -E -i "s/ServerPassword=\"[^\"]*\"/ServerPassword=\"\"/" /palworld/Pal/Saved/Config/LinuxServer/PalWorldSettings.ini
```
Dashboy1998, line 127:
```suggestion
elif [ -v SERVER_PASSWORD ]; then
    echo "No SERVER_PASSWORD"
    sed -E -i "s/ServerPassword=\"[^\"]*\"/ServerPassword=\"\"/" /palworld/Pal/Saved/Config/LinuxServer/PalWorldSettings.ini
```

I think we should remove SERVER_PASSWORD from the Dockerfile. 
If by default we have no variable named SERVER_PASSWORD then nothing will be overridden unless you explicitly state no password (With the elif -v). 

This way we don't always overwrite the password. I'm not worried about the extra write so much as removing a password someone has set in their file but not set as an environmental variable.

win5923: I also think this way is better. Thanks for your suggestion.

thijsvanloef:
@win5923 Will review tonight!

thijsvanloef (approved):
made some changes, but ready to merge

PR #306, opened 2024-02-07
Add arm64 compatibility

sonroyaalmerol:
## Context 

* Adds arm64 compatibility with a separate Docker tag and Dockerfile.

## Choices

* Added new gh actions build workflows for arm64 images.
* A different tag is required as a separate Dockerfile is needed for arm64 compatibility. Should also be better overall to avoid any breaking changes for current amd64 users.
* The arm64 Dockerfile uses a steamcmd base image with Box64 included for emulation. I had better performance with Box64 compared to FEX.
* rcon-cli is also built in the arm64 Dockerfile as they don't seem to include arm64-linux on their releases.
* start.sh has been slightly modified to generate a modified PalServer.sh script for arm64 and use Box64 to launch the server.

## Test instructions

1. See if gh actions successfully builds both amd64 and arm64 images.
2. Tested arm64 image in an Oracle ARM64 VPS Free Tier (4 CPU, 24 GB RAM)
3. Tested features in arm64 image (RCON, Server Commands, Backup, Restore, cronjobs, etc.)


sonroyaalmerol:
This should resolve #10 without needing much changes to the core scripts.

thijsvanloef:
@sonroyaalmerol Awesome work, made a few changes, if you could resolve the linting errors it will be ready to merge!

PR #315, opened 2024-02-07
Replace `:latest` uses with the specific latest supported version

Twinki14:
# Motivations
To prevent breaking-changes leaking into our suggested docker-compose and more importantly our helm-chart, we should be using the latest supported version of the docker image, not `:latest`

# Modifications
- Bump the helm-chart version to `0.24.1`, to match the docker-image version
- Replace `:latest`: with the true-latest supported version, `0.24.1`, in our example docker-compose, helm-chart, and k8s example

### DIFF
--- a/charts/palworld/Chart.yaml
+++ b/charts/palworld/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: palworld
-version: 0.24.1
+version: 0.24.2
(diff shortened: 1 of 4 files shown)


Dashboy1998:
Would we update the Readme too?

If so do we also update the other languages at each release or have the translators update the other languages?

PR #336, opened 2024-02-09
Changed array to map

Dashboy1998:
## Context 

[Docker Documentation](https://docs.docker.com/compose/compose-file/05-services/#environment) doesn't support `var="value"` but rather:
* `var: "value"`
* `var: value`
* `var=value`
It is worth noting that you can do `"var=value"` but it's not mentioned

* People keep using quotes in array declaration of environmental variables. Changes array to map which docker compose supports the use of quotes
* Currently `- SERVER_NAME="World of Pals"` appears as `"World of Pals"`
* Currently `- SERVER_NAME=World of Pals` appears as `World of Pals`

## Choices

* By switching to a map `SERVER_NAME: "World of Pals"` appears as `World of Pals`
*  `SERVER_NAME: World of Pals` appears as `World of Pals`
* Using maps also aligns our docker compose to be more align with k8s folder yamls as those tend to use maps (Ideally this would be in spell check somehow)

## Test instructions

1. Start the container
2. Verify using quotes does not add quotes around variables `docker exec -it palworld-server bash -c 'echo $SERVER_NAME'`

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,19 +72,19 @@ services:
         - 8211:8211/udp
         - 27015:27015/udp
       environment:
-         - PUID=1000
-         - PGID=1000
-         - PORT=8211 # Optional but recommended
-         - PLAYERS=16 # Optional but recommended
-         - SERVER_PASSWORD=worldofpals # Optional but recommended
-         - MULTITHREADING=true
-         - RCON_ENABLED=true
-         - RCON_PORT=25575
-         - TZ=UTC
-         - ADMIN_PASSWORD=adminPasswordHere
-         - COMMUNITY=false  # Enable this if you want your server to show up in the community servers tab, USE WITH SERVER_PASSWORD!
-         - SERVER_NAME=World of Pals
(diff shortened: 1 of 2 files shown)


thijsvanloef:
I have some doubts about changing arrays to map, as this moves away from the way the `.env` file and `docker run` expect the values.

Dashboy1998:
> .env

They'll never be able to be the same. While [Docker's documentation](https://docs.docker.com/compose/environment-variables/env-file/) suggests env can have quotes it appears to not be the case as the quotes appear in the values.

Docker run expects quotes with spaces (Our docker run doesn't actually run because of the spacing).

Docker runs supports `-e SERVER_DESCRIPTION="palworld-server-docker by Thijs van Loef"` and `-e "SERVER_DESCRIPTION=palworld-server-docker by Thijs van Loef" `

The yaml also supports `"SERVER_DESCRIPTION=palworld-server-docker by Thijs van Loef"`.

However the env does not support either.

thijsvanloef (approved):
LGTM

PR #337, opened 2024-02-09
Add an option to generate a WorldOption.sav file automatically

sonroyaalmerol:
## Context 

* A `WorldOption.sav` file is usually generated from a single player save and not a dedicated server. However, a dedicated server can still recognize the .sav file.
* Some settings are limited in dedicated servers such as PVP and `BaseCampWorkerMaxNum` being locked at 15. Having an option to generate a `WorldOption.sav` file would allow these settings to be modified.
* Currently, people are generating their `WorldOption.sav` files from this [website](https://palworldoptions.com/). While this works, it would negate the purpose of having the settings as env vars in a Docker container as having the `WorldOption.sav` file would ignore the `PalWorldSettings.ini` file generated from those env vars. (#293)

## Choices

* ~The way it works is a simple HTTP POST request with the compiled settings as its body to that same website which generates the .sav file and saved to the appropriate folder within the container. Another option is to have the tool itself within the Docker image but that would require us having Python running inside the container.~
* An internal Python script is executed to generate a JSON data from the compiled `PalWorldSettings.ini` file to be processed by [palworld-save-tools](https://github.com/cheahjs/palworld-save-tools) and generate the actual `WorldOption.sav` file.
* The script would delete the existing `WorldOption.sav` file and generate a new one every boot to make sure it follows the current values of the Docker env vars.
* GENERATE_WORLD_OPTION env var is added and is defaulted to false. Since the .sav file gets deleted every boot, having this env var to false would make sure that the `PalWorldSettings.ini` is used instead.
* If for some reason the ~POST request~ file generation fails, GENERATE_WORLD_OPTION is ignored and will revert back to using `PalWorldSettings.ini`.

## Test instructions

1. Started the container from scratch with default env vars and GENERATE_WORLD_OPTION=true. Check if it generates the .sav file (it shouldn't as the save folder does not exist at this point).
2. Restart the container after the save folder has been initialized by the dedicated server.
3. Check if it generates the .sav file.
4. Test if it follows the default configurations.
5. Modify env vars (ENABLE_PLAYER_TO_PLAYER_DAMAGE, IS_PVP, BASE_CAMP_WORKER_MAX_NUM, etc.)
6. Restart the container and see if the changes persist.
7. Set GENERATE_WORLD_OPTION to false.
8. Restart the container if it reverts back to the default configurations.


scripts/generate-worldoption.sh
```diff
@@ -0,0 +1,73 @@
...
+            echo "Generating WorldOption.sav done!"
+        else
+            echo "$error_msg: $http_status_code"
+            echo "$world_option_msg"
+        fi
+    else
+        echo "$no_directory_msg"
+        echo "$world_option_msg"
+    fi
+fi
```
Dashboy1998, line 73: That site is just a GUI for [palworld-worldoptions](https://github.com/legoduded/palworld-worldoptions). Why not just run that code directly?
```suggestion

```

Dashboy1998 (changes requested):

scripts/generate-worldoption.sh
sonroyaalmerol: I was thinking of doing that but the tool has 2 dependencies (Python and uesave-rs). Python will probably be fine but I don't think uesave-rs have pre-built binaries for arm64-linux. That would add another build step for the arm64 image (Rust is known for its loooong build times). If that's not a concern then yes I guess we could do that.

Dashboy1998: Don't the admin password and server password get sent to the web server too?

I'm not a fan of sending out passwords.

Dashboy1998: > I was thinking of doing that but the tool has 2 dependencies (Python and uesave-rs). Python will probably be fine but I don't think uesave-rs have pre-built binaries for arm64-linux. That would add another build step for the arm64 image (Rust is known for its loooong build times). If that's not a concern then yes I guess we could do that.

Should probably take a look at the couple time then for arm.

It would only be compiled when it's updated, otherwise it should just use the cached layer.

Diyagi: If you want an harder route but ditching the dependency on uesave, [palworld-save-tools](https://github.com/cheahjs/palworld-save-tools) run with only python and seems to be able to go from SAV > JSON and JSON > SAV. 

It presents the config options as GUIDs in the converted JSON file tho (probably bc it has no idea of what they are, since this is not supposed to be used with `WorldOption.sav`).

I did an quick test going from SAV to JSON, changing some values and going back to SAV, it seems to work.

MSpreckels:
Please also test this with existing savedata.

scripts/generate-worldoption.sh
sonroyaalmerol: > If you want an harder route but ditching the dependency on uesave, [palworld-save-tools](https://github.com/cheahjs/palworld-save-tools) run with only python and seems to be able to go from SAV > JSON and JSON > SAV.
> 
> I did an quick test going from SAV to JSON, changing some values and going back to SAV, it seems to work.

This seems like a good alternative! I'll have a look at this.

sonroyaalmerol:
I've removed the previous script of using the online API and opted to use the Python tools instead. I modified a portion of the [palworld-worldoptions](https://github.com/legoduded/palworld-worldoptions) script to generate a JSON file for the [palworld-save-tools](https://github.com/cheahjs/palworld-save-tools) to be usable instead of uesave-rs to remove the extra build step for arm64 images.

Tested it with both new and existing savedata and everything works as expected.

sonroyaalmerol:
Closing in favor of #380

PR #342, opened 2024-02-09
Add env to toggle to automatically generate palworldsettings

### LINKED ISSUE #326: Create switch to allow manual configuration of .ini files
**Is your feature request related to a problem? Please describe.**
The last update forced everyone to use environment files. I would like to see an option that allows users to turn automatic setting generation off if they would like.

**Describe the solution you'd like**
Setting an env var like MANUAL_SETTING_CONFIG=true to prevent the compile-setting.sh script from running

**Describe alternatives you've considered**
None

MSpreckels:
## Context 

resolves #326 

## Choices

Add an env var which is false by default that decides if the PalWorldSettings.ini is compiled from env vars automatically

#### Test Case: New Server, no automatic settings
1. Set `DISABLE_GENERATE_SETTINGS` to `true`
2. Start the server
3. On the server run `cat /palworld/Pal/Saved/Config/LinuxServer/PalWorldSettings.ini`
4. Verify that all values are default

#### Test Case: Existing Server, no automatic settings
1. Set `DISABLE_GENERATE_SETTINGS` to `true`
2. Modify or set `SERVER_NAME` environment variable
3. Start the server
4. On the server run `cat /palworld/Pal/Saved/Config/LinuxServer/PalWorldSettings.ini`
5. Verify that all values are still the same to prior launch and `SERVER_NAME` was not applied

#### Test Case: New Server, automatic settings
1. Set `DISABLE_GENERATE_SETTINGS` to `false` (or remove it entirely)
2. Start the server
3. On the server run `cat /palworld/Pal/Saved/Config/LinuxServer/PalWorldSettings.ini`
4. Verify that all values are as they were set as environment variables 

#### Test Case: Existing Server, automatic settings
1. Set `DISABLE_GENERATE_SETTINGS` to `false` (or remove it entirely)
2. Start the server
3. On the server run `cat /palworld/Pal/Saved/Config/LinuxServer/PalWorldSettings.ini`
4. Verify that all values are as they were set as environment variables 


scripts/start.sh
```diff
@@ -121,6 +117,30 @@ if [ "${MULTITHREADING,,}" = true ]; then
...
+      else
+          timeout --preserve-status 15s ./PalServer.sh 1> /dev/null
+      fi
+
+      # Wait for shutdown
+      sleep 5
+      cp /palworld/DefaultPalWorldSettings.ini /palworld/Pal/Saved/Config/LinuxServer/PalWorldSettings.ini
+  fi
+fi
+
```
Dashboy1998, line 143:
```suggestion
if [ "${DISABLE_GENERATE_SETTINGS,,}" = true ]; then
  printf "\e[0;32m%s\e[0m\n" "*****CHECKING FOR EXISTING CONFIG*****"

  # shellcheck disable=SC2143
  if [ ! "$(grep -s '[^[:space:]]' /palworld/Pal/Saved/Config/LinuxServer/PalWorldSettings.ini)" ]; then

      printf "\e[0;32m%s\e[0m\n" "*****GENERATING CONFIG*****"

      # Server will generate all ini files after first run.
      if [ "$architecture" == "arm64" ]; then
          timeout --preserve-status 15s ./PalServer-arm64.sh 1> /dev/null
      else
          timeout --preserve-status 15s ./PalServer.sh 1> /dev/null
      fi

      # Wait for shutdown
      sleep 5
      cp /palworld/DefaultPalWorldSettings.ini /palworld/Pal/Saved/Config/LinuxServer/PalWorldSettings.ini
else
  printf "\e[0;32m*****GENERATING CONFIGS*****\e[0m\n"
  /home/steam/server/compile-settings.sh
  fi
fi
```

I would change it so true is explicit. Otherwise if DISABLE_GENERATE_SETTINGS is anything other than false then it will generate. I don't think we want it to generate if DISABLE_GENERATE_SETTINGS is blank is set to say BANANAS.

Dashboy1998 (changes requested):

thijsvanloef (approved):
LGTM :)

PR #349, opened 2024-02-10
Update server without delay if nobody is connected

Luatan:
## Context 

* The Updater should not wait to update the server, if there are no Players connected.

## Choices

* To check how many players are connected we are using RCON, since there is currently no other method
* If players connects after the script has sent the message, the player will not know that there will be an update (if no discord webhook is used). This reduces the chance of this happening.

## Test instructions
1. set the following environment variables: 
```env
AUTO_UPDATE_ENABLED: true
AUTO_UPDATE_WARN_MINUTES: 1
AUTO_UPDATE_CRON_EXPRESSION: "* * * * *"
```
2. Connect to the Server with a client
3.  change the buildid in `palworld\steamapps\appmanifest_2394010.acf` to a different value
4. verify that the script waits for the time set in `AUTO_UPDATE_WARN_MINUTES`
5. wait for the server to restart
6. change the buildid again without connecting
7. verify that the server is being stopped immediately after the cronjob was ran.


raimannma:
We probably wanna combine this with my PR: https://github.com/thijsvanloef/palworld-server-docker/pull/347
Both need to get the playerlist

Dashboy1998:
> We probably wanna combine this with my PR: #347 Both need to get the playerlist

I was actually thinking we make a function in helper_functions called get_player_count.

Luatan:
> > We probably wanna combine this with my PR: #347 Both need to get the playerlist
> 
> I was actually thinking we make a function in helper_functions called get_player_count.

Should I create a separate PR for this or does someone else wanna do it?

raimannma:
> > > We probably wanna combine this with my PR: #347 Both need to get the playerlist
> > 
> > I was actually thinking we make a function in helper_functions called get_player_count.
> 
> Should I create a separate PR for this or does someone else wanna do it?

You could maybe just create a new PR with both our features implemented.
Or you could just copy my feature this pr and extract the players count to the helpers function.
Doesn't really matter to me :)

Luatan:
I have added the new helper function get_player_count. Now the auto_reboot.sh and update.sh script both use this function to check how many players are online.

thijsvanloef (approved):
Tested, LGTM

PR #353, opened 2024-02-10
feat: Publish to GHCR

USA-RedDragon:
## Context 

* Release the Docker images to the GitHub Container repository which is free and isn't subject to the same rate limiting as Docker Hub
* Related: #354

## Choices

* I added jobs in parallel with the Docker Hub release jobs to push to GHCR
* I added a workflow to publish developer images to GHCR as well

## Test instructions

1. Just ran the workflows in my fork, resulting in: https://github.com/USA-RedDragon/palworld-server-docker/pkgs/container/palworld-server-docker


Twinki14 (changes requested):
Instead of brand new workflows, we should keep docker image publishing all in the same workflow

Ideally this would be building the image then pushing to both registries in the same workflow job

USA-RedDragon:
Question, are you referring to the extra workflow for the GHCR Development Builds? Or do you mean I should be pushing to both GHCR and Docker Hub in the same job in `release.yaml`? (Or both?)

Twinki14 (approved):
Perfect! Could not have done it better myself

thijsvanloef (approved):

PR #364, opened 2024-02-11
feat: Unify Dockerfiles

USA-RedDragon:
## Context 

* Unify the Docker image for both amd64 and arm64 architectures so that arm64 users no longer need to use a separate tag. With this solution, no users will have to specify `-arm64` tags since Docker will resolve the appropriate architecture image automatically. 

## Choices

There's no real magic to this unification, it uses:
* [the Docker-provided `$TARGETARCH` build argument to choose which base image to use](https://github.com/thijsvanloef/palworld-server-docker/pull/364/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R15-R19)
* [opts to build `gorcon` for both arm64 and amd64](https://github.com/thijsvanloef/palworld-server-docker/pull/364/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R1-R13)
* [swaps for the `arm64v8/golang:1.22.0-bullseye` image for official `golang` alpine image when building `gorcon`](https://github.com/thijsvanloef/palworld-server-docker/pull/364/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R1)
* [disables CGO for `gorcon` compilation so that the GCC compiler toolchain doesn't have to be installed and the binary is portable](https://github.com/thijsvanloef/palworld-server-docker/pull/364/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R7)
* [removes the deprecated `go get` command since `go build` will resolve any missing dependencies based on the `go.sum` file](https://github.com/thijsvanloef/palworld-server-docker/pull/364/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R13)
* [uses the  `$TARGETARCH` build argument to choose which SHA1SUM to check against per-architecture](https://github.com/thijsvanloef/palworld-server-docker/pull/364/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R54-R57)
* [Removes `--progress=dot:giga` from `wget` in the Go build of `gorcon` since the Alpine/Busybox variant of `wget` doesn't support this option, but utilizes `-q` to to comply with Hadolint DL3047](https://github.com/thijsvanloef/palworld-server-docker/pull/364/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R8)
* [I did not unify the `unit-test.yml` workflow as doing so would prevent running the arm64 and amd64 tests in parallel like they currently are.](https://github.com/thijsvanloef/palworld-server-docker/pull/364/files#diff-d4223b44df678cece85f2542c19b9cd060d43b22f753c25214d5942ef2ca5d18R118)
* [Ignores Hadolint DL3029 (`Do not use --platform flag with FROM`) as on amd64 this image must still resolve to avoid build errors but is not used in the final image, but it is used in the arm64 build.](https://github.com/thijsvanloef/palworld-server-docker/pull/364/commits/fa167c5f5ef5e5e336902dfc85e9b02d29c34d40)
* [Ignores Hadolint DL3006 (`Always tag the version of an image explicitly`) as Hadolint is unaware that the final FROM image is based on one of the two FROM's above, both of which are tagged. Therefore this is a false positive.](https://github.com/thijsvanloef/palworld-server-docker/pull/364/commits/f0d780179133529eef3a977738443115e72ec90f)
* and basically cuts the CI workflow YAMLs in half

As @sonroyaalmerol pointed out in https://github.com/thijsvanloef/palworld-server-docker/pull/319#issuecomment-1937757999, this could possibly require building future dependencies from source if added, though the `arm64` emulation in GitHub Actions is likely to always be the slowest point in the image build. Using `buildx` now means that some of the workflow can run in parallel vs the default builder which was sequential, meaning it's likely the time to build has been reduced by a bit.

## Test instructions

1. Built for both architectures with `docker buildx build --platform=linux/amd64,linux/arm64 .` to verify build sanity
2. On my amd64 desktop, `docker buildx build --platform=linux/amd64 . --load -t palworld:test`
3. (still on amd64 desktop) `docker run -p 8211:8211/udp --rm -it palworld:test` and connected via Palworld to verify the server still runs
4. On a Macbook Air M2, I disabled Rosetta in the Docker for Mac settings
5. (still on Macbook) `docker buildx build --platform=linux/arm64 . --load -t palworld:test`
6. (still on Macbook)  `docker run -p 8211:8211/udp --rm -it palworld:test` and connected via Palworld to verify the server still runs

I have introduced breaking changes in that the `-arm64` tags suffixes will no longer be required.


USA-RedDragon:
Hadolint makes two good points that I don't want to ignore:

> `Dockerfile:18 DL3007 warning: Using latest is prone to errors if the image will ever update. Pin the version explicitly to a release tag` 
This is the `sonroyaalmerol/steamcmd-arm64:latest` base image. I personally solved for this by pinning the `sha256` digest ([here](https://github.com/USA-RedDragon/palworld-server-docker/blob/next/Dockerfile#L35)), but I also utilize [Renovate](https://github.com/apps/renovate) to get PRs when this changes

> `Dockerfile:41 DL3002 warning: Last USER should not be root`
Which is solid advice in general, I would consider creating and using appropriate users to be out of scope for this PR, but I'd be happy to make a separate PR for this if y'all want it!

Dockerfile
```diff
@@ -1,11 +1,45 @@
...
+    && tar -xzvf rcon.tar.gz \
+    && rm rcon.tar.gz \
+    && mv rcon-cli-${RCON_VERSION}/* ./ \
+    && rm -rf rcon-cli-${RCON_VERSION} \
+    && go build -v ./cmd/gorcon
+
+FROM cm2network/steamcmd:root as base-amd64
+# Ignoring --platform=arm64 as this is required for the multi-arch build to continue to work on amd64 hosts
+# hadolint ignore=DL3029
+FROM --platform=arm64 sonroyaalmerol/steamcmd-arm64:latest as base-arm64
```
sonroyaalmerol, line 18: One thing you could do is use `sonroyaalmerol/steamcmd-arm64:root` instead of the `latest` tag. This would also allow you to remove `USER root` from the other linting error. While this is definitely not a fix, at least you'll have the linting errors suppressed for now.

USA-RedDragon: Oh hey, that's you!

What's the difference between the `root` and `latest` tags? I noticed that the `arm64` build used `latest`, so I was a bit leery of just swapping tags.

Though that would avoid the lint error, the "spirit" of the lint rule would still be violated, as it's point is to avoid tags that aren't deterministic

sonroyaalmerol: The image is identical to `cm2network/steamcmd` (the base image used for amd64) but with the addition of Box64 and Box86 which adds compatibility for arm64. The `root` tag is literally just the `latest` tag but with `root` as its active user. The `latest` tag has `steam` as its active user as the steamcmd executable within the images are owned and can only be executed by `steam`.

The only reason the `Dockerfile.arm64` used the `latest` tag was that I hadn't implemented the `root` tag yet when I PR-ed the arm64 support.

Well, I can definitely add some deterministic tags for version pinning in my image but I can't say the same for the original `cm2network/steamcmd` image.

USA-RedDragon: Thanks for clarifying!

win5923 (approved):

win5923:
Excellent！

PR #378, opened 2024-02-12
Wait for the server to shutdown before restoring a backup

### LINKED ISSUE #359: Restoring backup from another server does not work using restore script
## Describe the bug

Upon restoring a backup file from another server using `docker exec -it palworld-server restore`, the hash in `/palworld/Pal/Saved/Config/LinuxServer/GameUserSettings.ini` remains unchanged. Thus game loads with incorrect save state.

I presume that backup and restore works for backups made on the same server since the hash would remain the same.

## To Reproduce

1. Create new server through Docker
2. Copy backup file from another server into backup folder of restore server
3. Run `docker exec -it palworld-server restore` and select correct backup file, answer 'y'
4. view GameUserSettings.ini file to see that hash has not changed and is different from the config file from the backup file when manually extracted.
5. Connect to game from Palworld game, the game prompts to setup game character. (note that while the restore script instructions says to restart server manually, the server has been restarted already.

## Expected behavior

Restore from another server should bring up the saved game state of that server. GameUserSettings.ini should reflect this.

## Actual behavior

Save state of the other server is not restored. Files have been copied and looks to be present. However, GameUserSettings.ini's reference to the save directory hash is incorrect.

Upon looking into the issue, we can see that the backup config file has indeed been copied. However, it gets modified shortly after.
image

Looking into the script a bit more, it looks like the container restarts after server shutdown command `rcon-cli -c /home/steam/server/rcon.yaml "shutdown 1"`.

What appears to be happening is that the game shutdown might be writing to the config file upon shutdown. This likely means that in order for this to work, there would need to be two options:
1. rewrite the restore script as an external data restore tool
2. change how the container starts the palworld server such that palworld server shutdown does not affect the container's operation. This would allow the backup to be restored, and then the backup script should attempt to start palworld server again.

## OS information

- OS:  Ubuntu 22.04.2
- Palworld Version:  0.1.4.1
- Docker Image Version:  sha256:4ed9eef7213960d03e752d8f65c75f22e69e54a6353da3fa01d0326b2450b1c3

## Hardware information

- CPU:  i7-13700k (source server), i3-2500T (test server)
- Memory:  32GB

## docker-compose.yml contents

I'm using ansible. It's pretty easy to read so I hope no one minds if I just paste it instead of writing as docker-compose.

```yaml
- name: Setup Service - Palworld Server (docker)
  community.docker.docker_container:
    name: palworld-server
    hostname: "{{ ansible_hostname }}"
    image: thijsvanloef/palworld-server-docker:latest
    pull: true
    ports:
      - 8211:8211/udp
      - 27015:27015/udp
    volumes:
      - "{{ palworld_data_dir }}:/palworld"
    restart_policy: unless-stopped
    env:
      TZ: "America/Toronto"
      PUID: '997'
      PGID: '997'
      PORT: '8211'
      PLAYERS: "{{ palworld_setting_players }}"
      SERVER_PASSWORD: "{{ palworld_server_password }}"
      MULTITHREADING: 'true'
      RCON_ENABLED: 'true'
      RCON_PORT: '25575'
      ADMIN_PASSWORD: "{{ palworld_admin_password }}"
      COMMUNITY: 'false'
      SERVER_NAME: "{{ palworld_setting_server_name }}"
    cgroup_parent: "{{ palworld_cgroups_parent }}"
    cpuset_cpus: "{{ palworld_cpuset_cpus }}"
    memory: "{{ palworld_memory }}"
    memory_swappiness: "{{ palworld_memory_swappiness }}"
```

## Container/Host Logs

I haven't had time to review if it has any sensitive information so will refrain from uploading. However, based on what I've seen, I don't think it's necessary for troubleshooting.

Luatan:
## Context 

* The restore script does not wait for the server to actually shut down
* Because of this Settings are reverted after the server restarts
* Closes #359 

## Choices

* Getting the PID of the PalServer-Linux-Test binary to test if it is still running
* Continue with the restore only after the server has shutdown

## Test instructions

1. Start a Server
2. Create a backup
3. Extract the backup and change the id in `/palworld/Pal/Saved/Config/LinuxServer/GameUserSettings.ini`
4. Rename the directory in `/palworld/Pal/Saved/SaveGames/0/` to the ID in GameUserSettings.ini
5. Tar the 'Saved' directory again
6. Run the restore command `docker exec -it palworld-server restore`
7. Select the backup which was altered
8. Verify that the server has restarted
9. Verify that the ID in GameUserSettings.ini is the same which was used in the backup
10. Verify that the directory in `/palworld/Pal/Saved/SaveGames/0/` has the correct name

### DIFF
--- a/scripts/restore.sh
+++ b/scripts/restore.sh
@@ -109,6 +109,7 @@ if [ -f "$BACKUP_FILE" ]; then
             # Decompress the backup file in tmp directory
             tar -zxvf "$BACKUP_FILE" -C "$TMP_PATH"
 
+            # Make sure Saves with a different ID are removed before restoring the save
             rm -rf "$RESTORE_PATH/Saved/"
 
             # Move the backup file to the restore directory
--- a/scripts/restore.sh
+++ b/scripts/restore.sh
@@ -67,6 +67,12 @@ if [ -f "$BACKUP_FILE" ]; then
             echo "RCON is not enabled. Please enable RCON to use this feature. Unable to restore backup."
             exit 1
         fi
+
+        while [ -n "$(pidof PalServer-Linux-Test)" ] 
+        do
+            echo "Waiting for Palworld to exit.."
+            sleep 1
+        done
         printf "\e[0;32mShutdown complete.\e[0m\n"
 
         trap - ERR
@@ -102,6 +108,8 @@ if [ -f "$BACKUP_FILE" ]; then
             
             # Decompress the backup file in tmp directory
             tar -zxvf "$BACKUP_FILE" -C "$TMP_PATH"
+
+            rm -rf "$RESTORE_PATH/Saved/"
 
             # Move the backup file to the restore directory
             \cp -rf -f "$TMP_PATH/Saved/" "$RESTORE_PATH"


scripts/restore.sh
```diff
@@ -67,6 +67,12 @@ if [ -f "$BACKUP_FILE" ]; then
             echo "RCON is not enabled. Please enable RCON to use this feature. Unable to restore backup."
             exit 1
         fi
+
+        while [ -n "$(pidof PalServer-Linux-Test)" ] 
+        do
+            echo "Waiting for Palworld to exit.."
+            sleep 1
+        done
```
Dashboy1998, line 75:
```suggestion
        mapfile -t server_pids < <(pgrep PalServer-Linux-Test)
        if [ "${#server_pids[@]}" -ne 0 ]; then
            echo "Waiting for Palworld to exit.."
            for pid in "${server_pids[@]}"; do
                tail --pid="$pid" -f 2>/dev/null
            done
        fi
```

I would remove the need for sleep using tail. 
Also I would use an array so in case there's multiple running.
Code is essentially what's in init.sh.

scripts/restore.sh
```diff
@@ -103,6 +109,9 @@ if [ -f "$BACKUP_FILE" ]; then
             # Decompress the backup file in tmp directory
             tar -zxvf "$BACKUP_FILE" -C "$TMP_PATH"
 
+            # Make sure Saves with a different ID are removed before restoring the save
+            rm -rf "$RESTORE_PATH/Saved/"
+
```
Dashboy1998, line 114: Instead of deleting the directory should we instead change `\cp -rf "$RESTORE_PATH/Saved" "$TMP_SAVE_PATH/Saved"` to `mv "$RESTORE_PATH/Saved" "$TMP_SAVE_PATH/Saved"`?

Dashboy1998 (changes requested):

scripts/restore.sh
Luatan: I tried your suggestion, but I get an Error if we only use mv 
```
mv: cannot move '/palworld/backups/tmp.D0rNX7pZLE/Saved/' to '/palworld/Pal/Saved': Directory not empty
An error occurred during restore.
```

Dashboy1998: I'm guessing you did line 118?

I meant 96.

Although let's keep it simple and just use the rm you added. I'd like to the the issue as have fixed before we try to change a cp to mv because that would require more testing.

Dashboy1998 (approved):

scripts/restore.sh
Luatan: Oww.., okay. Thought you meant the line 118. You're right, that would work. My bad.

PR #380, opened 2024-02-13
Add an option to generate a WorldOption.sav file automatically in (mostly) bash

sonroyaalmerol:
## Context 

* A `WorldOption.sav` file is usually generated from a single player save and not a dedicated server. However, a dedicated server can still recognize the .sav file.
* Some settings are limited in dedicated servers such as PVP and `BaseCampWorkerMaxNum` being locked at 15. Having an option to generate a `WorldOption.sav` file would allow these settings to be modified.
* Currently, people are generating their `WorldOption.sav` files from this [website](https://palworldoptions.com/). While this works, it would negate the purpose of having the settings as env vars in a Docker container as having the `WorldOption.sav` file would ignore the `PalWorldSettings.ini` file generated from those env vars. (#293)

## Choices
* Unlike the previous PR (#337), this minimizes the Python code and removes the unnecessary parts for better maintainability. However, the script assumes that the generated `PalWorldSettings.ini` is always valid.
* This script generates the `WorldOption.sav` file by:
  * Parsing configuration settings from a `PalWorldSettings.ini` file.
  * Converting the configuration settings into a JSON format suitable for [palworld-save-tools](https://github.com/cheahjs/palworld-save-tools).
  * Incorporating the generated JSON into the provided `WorldOption.json.template`.
  * Using an embedded Python script with [palworld-save-tools](https://github.com/cheahjs/palworld-save-tools) to compress the JSON data into a `WorldOption.sav` file.
  * Writing the resulting .sav file to the appropriate directory within the game server's file structure.
* The script would delete the existing `WorldOption.sav` file and generate a new one every boot to make sure it follows the current values of the Docker env vars.
* `GENERATE_WORLD_OPTION` env var is added and is defaulted to false. Since the .sav file gets deleted every boot, having this env var to false would make sure that the `PalWorldSettings.ini` is used instead.

## Test instructions
New server test:
1. Started the container from scratch with default env vars and GENERATE_WORLD_OPTION=true. Check if it generates the .sav file (it shouldn't as the save folder does not exist at this point).
2. Restart the container after the save folder has been initialized by the dedicated server.
3. Check if it generates the .sav file.
4. Test if it follows the default configurations.
5. Modify env vars (ENABLE_PLAYER_TO_PLAYER_DAMAGE, IS_PVP, BASE_CAMP_WORKER_MAX_NUM, etc.)
6. Restart the container and see if the changes persist.
7. Modify env vars back to defaults.
8. Restart the container and see if the changes persist.
9. Set GENERATE_WORLD_OPTION to false.
10. Restart the container if it reverts back to using `PalWorldSettings.ini`.

Existing server test:
1. Started an existing container with BASE_CAMP_WORKER_MAX_NUM=20 (from the default 15) and GENERATE_WORLD_OPTION=true.
2. Check if it generates the .sav file.
3. Test if it follows the set configurations.
4. Modify more env vars (ENABLE_PLAYER_TO_PLAYER_DAMAGE, IS_PVP, etc.)
5. Restart the container and see if the changes persist.
6. Modify env vars back to defaults.
7. Restart the container and see if the changes persist.
8. Set GENERATE_WORLD_OPTION to false.
9. Restart the container if it reverts back to using `PalWorldSettings.ini`.


sonroyaalmerol:
Hi @gzukel, do you mind sharing the log with DEBUG env var set to true?

scripts/generate-worldoption.sh
```diff
@@ -0,0 +1,208 @@
...
+    local output_path="$2"
+
+    echo "WorldOption Generator: Compressing WorldOption to .sav"
+    if echo "$json_data" | python3 -c "$convert_json_to_sav_python" > "$output_path/WorldOption.sav"; then
+        echo "WorldOption Generator: Generated WorldOption.sav file to $output_path"
+        echo "Generating WorldOption.sav done!"
+    else
+        echo "WorldOption Generator: WorldOption.sav generation failed."
+    fi
+}
```
Dashboy1998, line 153: Please don't include python code in this way.
Make a separate python file and call that.

Dashboy1998 (changes requested):

sonroyaalmerol:
Closing PR due to its fragility and complexity. Will wait until I can think of a solution without adding an additional language to maintain before opening up another PR as we prefer to currently do things strictly in bash.

Issue #358, opened 2024-02-11
`isWritable` check requires that the root directory be deletable ("/palworld is not writable")


acmcarther:
## Describe the bug

When I updated today, my server would not start with error "/palworld is not writable"

I root caused this [to this script](https://github.com/thijsvanloef/palworld-server-docker/blob/8fdc434c21e144ca73d5b60b64b2bf8eb551898e/scripts/helper_functions.sh#L49-L52), which is checking not only for writability, but also for the ability to delete the directory.

I suppose due to some idiosyncracy of my Kubernetes setup (NFS based volume management via nfs-client-provisioner), my persistent volume claim root directories are not created to be deletable. Normally an application would not try to delete the whole PVC directory anyway, so this has historically not been a problem.

## To Reproduce

1. Create a palworld directory without delete permission on the root palworld directory
2. Startup
3. "/palworld is not writable"

## Expected behavior

Palworld should not require the root palworld directory to be deletable ideally since there's no circumstance where that directory would need to be deleted.

Alternatively, the file writability check could be for some file *within* the palworld world directory

Alternatively (2), the error message could be more explicit about the required permissions.

---

I resolved this locally by manually updating the NFS directory to be deletable, but the kubernetes nfs-client-provisioner does not normally create the directory this way, so I'll have to do it manually for future palworld servers.

Dashboy1998:
Very interesting.

What are you running the NFS server on?

When I had made the write check it assumed Linux permissions. Not sure how it works for more advance permissions like AppArmor or SELinux.

I agree it shouldn't need the ability to delete the directory. 

Checking if a child directory or file is writable wouldn't give the result we want. I suppose we could try to create a temporary file inside palworld directory.

I'm not sure how we would be more explicit as it sounds like you are using a different permissions set than base Linux. 

I'll check if there's another way to check if it's writable.

muri235:
I also geht the same error. But even setting the root directory and the one above to chmod 777 doesn't change anything. Still getting the "/palworld is not writable" eventhough "/palworld/backup/" is being created.

XeNz:
Also getting the same issue.
Not using NFS but ext4.
Folder has 777 permissions.

Dashboy1998:
@XeNz @muri235 Are you both using Kubernetes?

Dashboy1998:
> I am not running on k8s. I'm running a simple docker compose file.

What distro are you running?
Are you using apparmor or selinux?

[...]

PR #381, opened 2024-02-13
Updated isWritable

Dashboy1998:
## Context 

* Addressing #358 

## Choices

* Since `[ -w path ]` won't work if a directory is writable but not delete-able we check if it is a directory and if so then try to create a temporary file.

## Test instructions

1. Start normally. If this works then isWritable works for files and directories that are writable.
2. Verify the function still works for files that are not writable: `chmod 444 palworld/Pal/Saved/Config/LinuxServer/PalWorldSettings.ini`
3. Verify the function still works for directories that are writable and delete-able: `chmod 444 palworld/Pal/Saved/Config/LinuxServer/`
4. Verify the function works for directories that are writable but not delete-able


Dashboy1998:
@acmcarther can you test if this resolves your issue? I am unable to test as I do not have an NFS.

Issue #382, opened 2024-02-13
Unable to build AMD64 image with versions locked in


Dashboy1998:
## Describe the bug

I am unable to build the image with the current `apt-get install package-version` as the version seems to be breaking it.

## To Reproduce

1. Attempt to build

## Expected behavior

Should build

## Actual behavior

```
=> ERROR [palworld stage-3 1/6] RUN apt-get update && apt-get install -y   1.7s
------                                                                           
 > [palworld stage-3 1/6] RUN apt-get update && apt-get install -y --no-install-recommends     procps=2:3.3.17-5     wget=1.21-1+deb11u1     gettext-base=0.21-4     xdg-user-dirs=0.17-2     jo=1.3-2     && apt-get clean     && rm -rf /var/lib/apt/lists/*:
0.223 Get:1 http://deb.debian.org/debian bookworm InRelease [151 kB]
0.262 Get:2 http://deb.debian.org/debian bookworm-updates InRelease [52.1 kB]
0.275 Get:3 http://deb.debian.org/debian-security bookworm-security InRelease [48.0 kB]
0.301 Get:4 http://deb.debian.org/debian bookworm/main amd64 Packages [8786 kB]
0.429 Get:5 http://deb.debian.org/debian bookworm-updates/main amd64 Packages [12.7 kB]
0.429 Get:6 http://deb.debian.org/debian-security bookworm-security/main amd64 Packages [137 kB]
0.960 Fetched 9187 kB in 1s (11.6 MB/s)
0.960 Reading package lists...
1.271 Reading package lists...
1.595 Building dependency tree...
1.690 Reading state information...
1.694 Package wget is not available, but is referred to by another package.
1.694 This may mean that the package is missing, has been obsoleted, or
1.694 is only available from another source
1.694 
1.694 Package procps is not available, but is referred to by another package.
1.694 This may mean that the package is missing, has been obsoleted, or
1.694 is only available from another source
1.694 However the following packages replace it:
1.694   libproc2-0
1.694 
1.694 Package gettext-base is not available, but is referred to by another package.
1.694 This may mean that the package is missing, has been obsoleted, or
1.694 is only available from another source
1.694 However the following packages replace it:
1.694   libasprintf0v5
1.694 
1.694 Package xdg-user-dirs is not available, but is referred to by another package.
1.694 This may mean that the package is missing, has been obsoleted, or
1.694 is only available from another source
1.694 
1.694 Package jo is not available, but is referred to by another package.
[...]

PR #384, opened 2024-02-13
Removed package version

Dashboy1998:
## Context 

* Fix issues when packages are removed from repos causing missing versions #382 

## Choices

* Most packages we're installing are fairly stable so there's not much reason to force a version.

## Test instructions

1. Test that is build


thijsvanloef:
Do not agree, let's just change the package versions to the ones that are available. Not pinning the versions introduces more variables into why the server could not work

Dashboy1998:
> Do not agree, let's just change the package versions to the ones that are available. Not pinning the versions introduces more variables into why the server could not work

Might run into issues later on then with the ARM image if the repos don't have matching versions.

PR #385, opened 2024-02-13
update package versions

thijsvanloef:
## Context 

Docker Builds are failing as cm2network/steamcmd:root got updated, this should solve the package errors for the adm64,
@sonroyaalmerol Needs to update his image as well to fix the issues for arm

## Choices

* 

## Test instructions

1. 


sonroyaalmerol:
I just switched to the bookworm base image as well so it should be fine now. Thanks for the heads up!

thijsvanloef:
Sadly for now we need to not pin wget

PR #356, opened 2024-02-11
Introduced Utility Functions - Remove duplicate code snippets 

Phelms215:
## Context 

* Introduced utility function for Logging, SteamCMD and Discord Messaging 
* Remove redundant (copy & paste) code snippets 
* Remove nesting where possible
* Better Update/Install required logic
* Removed requirement for UPDATE_ON_BOOT for initial installation 

## Choices
* Clean up redundant code to make future enhancements easier to implement 
* Removed requirement for UPDATE_ON_BOOT if server installation is not detected (no PalServer.sh or appmanifest file) 

## Test instructions

1. Started with old server version, ran through reboots with UPDATE_ON_BOOT and AUTOUPDATE flags set
2. Started new server and validated installation
3. Tested backup script 
4. Tested discord messaging 
5. Tested logging 

## Checklist before requesting a review

* [X] I have performed a self-review of my code
* [ ] I've added documentation about this change to the README.
* [X] I've not introduced breaking changes.

PR #392, opened 2024-02-15
Move countdown to function for auto update and auto reboot

Dashboy1998:
## Context 
* Due to all the changes in #356 this now has the features for/from the following
  -   #395
  -   #398 
  -   #400
  -   #393 
  - Removed use of `$?` in favor of `if command`

* Auto update outputs a single message to the server saying it will update in N minutes whereas auto reboot outputs a message counting down N minutes.

## Choices

* Changed to match behavioral of auto reboot.
* Due to the massive merge of #356 I had to implement other fixes.

### Auto reboot

- Set AUTO_REBOOT_CRON_EXPRESSION close to your current time
- Set AUTO_REBOOT_ENABLED to true
- Set AUTO_REBOOT_EVEN_IF_PLAYERS_ONLINE to true
- Set AUTO_REBOOT_WARN_MINUTES to 5

#### Players in game and messages
Join the server and verify that the server restarts in 5 minutes after the cronjob has started and you receive messages for 5, 2, 1 minutes, and 30, 10 seconds.

#### Players leave game during countdown
Join the server and once the cronjob starts leave the game and verify it restarts on the minute after you left.

### Auto Update

- Set UPDATE_CRON_EXPRESSION close to your current time
- Set AUTO_UPDATE_ENABLED to true
- Set AUTO_REBOOT_WARN_MINUTES to 5
- Set UPDATE_ON_BOOT to true
- After the server starts change the buildId on line 33 in the file `palworld\steamapps\appmanifest_2394010.acf`` to any other value (for example 63707356556294349890)

#### Players in game and messages
Join the server and verify that the server restarts in 5 minutes after the cronjob has started and you receive messages for 5, 2, 1 minutes, and 30, 10 seconds.

#### Players leave game during countdown
Join the server and once the cronjob starts leave the game and verify it restarts on the minute after you left.


Luatan:
> ## Test instructions
> 1. Set UPDATE_CRON_EXPRESSION to */5 * * *

needs to be "*/5 * * * *"

> 2. Set WARN_MINUTES to 2
> 3. check in the log that the cronjob is running correctly
> 4. change the buildId on line 11 in the file `palworld\steamapps\appmanifest_2394010.acf`` to any other value (for example 13289561)

I think you mean to change the manifestid on line 32 right?

Then it works 😊

Do we really want to send an notification every minute? I think that's a bit too many if the default is 30 minutes.
My suggestion would be to have set intervals. For example:
30 15 10 5 2 1

And if the auto_reboot and auto_update use this, I think it would also be a good idea to make a helper function for it.

Dashboy1998:
> > ## Test instructions
> > 1. Set UPDATE_CRON_EXPRESSION to */5 * * *
> 
> needs to be "*/5 * * * *"
> 
> > 2. Set WARN_MINUTES to 2
> > 3. check in the log that the cronjob is running correctly
> > 4. change the buildId on line 11 in the file `palworld\steamapps\appmanifest_2394010.acf`` to any other value (for example 13289561)
> 
> I think you mean to change the manifestid on line 32 right?
> 
> Then it works 😊
> 
> Do we really want to send an notification every minute? I think that's a bit too many if the default is 30 minutes.
> My suggestion would be to have set intervals. For example:
> 30 15 10 5 2 1
> 
> And if the auto_reboot and auto_update use this, I think it would also be a good idea to make a helper function for it.

Ah the mistakes when I copy the instructions from another PR.

I just did every minute because that's what the auto reboot does.

I can add a case statement so it's not every minute.

scripts/auto_reboot.sh
Luatan, line 1: What I think is missing is that it only should countdown, if there are players connected. If no players are connected, the countdown doesn't really make sense in my opinion.

Dashboy1998: I've actually been working on that functionality.

Been also trying to get it to skip the countdown once no players are connected. Right now it only checks in the beginning if players are connected, trying to make it so even while counting down it checks if there are no players so if it is now empty it skips.

Luatan: That's a good idea yeah.

scripts/helper_functions.sh
```diff
@@ -93,96 +96,192 @@ export YellowBoldText='\033[1;33m'      # Yellow
...
+                esac
+            fi
+        done
+        return 0
+    else
+        # mtime is not an integer so check if there are no players
+        if [ "$(get_player_count)" -gt 0 ]; then
+            return_val=1
+        fi
+    fi
```
Luatan, line 285: I would suggest to negate `if [[ "${mtime}" =~ ^[0-9]+$ ]]; then` to reduce nested code.

scripts/update.sh
```diff
@@ -2,32 +2,33 @@
...
+    fi
 
-if [ "${RCON_ENABLED,,}" = false ]; then
-    LogWarn "An update is available however auto updating without rcon is not supported"
-    DiscordMessage "An update is available however auto updating without rcon is not supported" "warn"
-    exit 1
-fi
+    if [[ "${AUTO_UPDATE_WARN_MINUTES}" =~ ^[0-9]+$ ]]; then
+        DiscordMessage "Server will update in ${AUTO_UPDATE_WARN_MINUTES} minutes"
+    fi
```
Luatan, line 21: I think we would need to cancel the update if WARN_MINUTES is not an integer.
```suggestion
    if [[ "${AUTO_UPDATE_WARN_MINUTES}" =~ ^[0-9]+$ ]]; then
        LogError "AUTO_UPDATE_WARN_MINUTES is not an integer: ${AUTO_UPDATE_WARN_MINUTES}. Please check your environment variables"
        DiscordMessage "AUTO_UPDATE_WARN_MINUTES is not an integer: ${AUTO_UPDATE_WARN_MINUTES}. Please check your environment variables" "failure"
        exit 1
    fi
    DiscordMessage "Server will update in ${AUTO_UPDATE_WARN_MINUTES} minutes"
```

scripts/update.sh
```diff
@@ -2,32 +2,33 @@
...
-rm /palworld/steamapps/appmanifest_2394010.acf
-backup
-RCON "shutdown 1"
\ No newline at end of file
+        backup
+        shutdown_server
+    elif [ -z "${AUTO_UPDATE_WARN_MINUTES}" ]; then
+        LogWarn "Unable to auto update, the server is not empty and AUTO_UPDATE_WARN_MINUTES is empty."
+    else
+        LogWarn "Unable to auto update, the server is not empty and AUTO_UPDATE_WARN_MINUTES is not an integer: ${AUTO_UPDATE_WARN_MINUTES}"
```
Luatan, line 32: Can be removed if the check for an Integer is already made above.

scripts/start.sh
```diff
@@ -78,27 +73,27 @@ if [ "${MULTITHREADING,,}" = true ]; then
...
-      else
-          timeout --preserve-status 15s ./PalServer.sh 1> /dev/null
-      fi
-
-      # Wait for shutdown
-      sleep 5
-      cp /palworld/DefaultPalWorldSettings.ini /palworld/Pal/Saved/Config/LinuxServer/PalWorldSettings.ini
-  fi
+    LogAction "GENERATING CONFIG"
+    LogWarn "Env vars will not be applied due to DISABLE_GENERATE_SETTINGS being set to TRUE!"
```
Luatan, line 77:
```suggestion
    LogWarn "Environment variables will not be applied due to DISABLE_GENERATE_SETTINGS being set to TRUE!"
```

scripts/auto_reboot.sh
```diff
@@ -2,32 +2,25 @@
...
-    RCON "shutdown 1"
-    exit 0
+    if countdown_message "${AUTO_REBOOT_WARN_MINUTES}" "The_Server_will_reboot"; then
+        shutdown_server
+    elif [ -z "${AUTO_REBOOT_WARN_MINUTES}" ]; then
+        LogError "Unable to auto reboot, the server is not empty and AUTO_REBOOT_WARN_MINUTES is empty"
+        exit 1
+    else
+        LogError "Unable to auto reboot, the server is not empty and AUTO_REBOOT_WARN_MINUTES is not an integer: ${AUTO_REBOOT_WARN_MINUTES}"
+    fi
```
Luatan, line 22: I think you can do these checks before you start the countdown. Would make it a bit cleaner in my opinion.

Luatan:
>(https://github.com/thijsvanloef/palworld-server-docker/pull/356) this now has the features for/from the following
> 
> * [Use DoExit instead of Shutdown #395](https://github.com/thijsvanloef/palworld-server-docker/issues/395)
> * [Remove the need of restart policy for auto update and auto reboot #399](https://github.com/thijsvanloef/palworld-server-docker/issues/399)
>

As far as I can tell a reboot policy is is still needed for the container both reboot and update. right?

scripts/update.sh
Dashboy1998: I have it like that because if you so choose you could run the update script with no one in the server and it will update.

Also if it is a cron job and there is no one why not just update? The timer is supposed to just let players know and not prevent the update as that's what the enable flag is for

scripts/helper_functions.sh
Dashboy1998: I have that because the only time this is called is during a shutdown for update or reboot.
I believe both of those should happen regardless if there's a timer or not as the timer is just to let players know. If there is no timer set and there are no players I believe the action should proceed.

scripts/update.sh
Dashboy1998: No, the integer check is different from this.

The integer checks only happens if there's an issue with the countdown, since the countdown is set to not check the timer if there are no players the second elif only happens if there are players and the countdown is empty or is not an integer.

The two LogWarn give you different information.

scripts/auto_reboot.sh
Dashboy1998: I have it like that because if you so choose you could run the reboot script with no one in the server and it will reboot.

Also if it is a cron job and there is no one why not just reboot? The timer is supposed to just let players know and not prevent the reboot as that's what the enable flag is for

Dashboy1998:
> Nice work!
> 
> > (#356) this now has the features for/from the following
> > 
> > * [Use DoExit instead of Shutdown #395](https://github.com/thijsvanloef/palworld-server-docker/issues/395)
> > * [Remove the need of restart policy for auto update and auto reboot #399](https://github.com/thijsvanloef/palworld-server-docker/issues/399)
> 
> As far as I can tell a reboot policy is is still needed for the container both reboot and update. right?

Yes.

scripts/update.sh
Luatan: I see what you mean. For me it just seems like we are allowing the user to misconfigure the env variables, which feels wrong. I think we should tell the user to fix it, but maybe we should do that in the start.sh then. To do it immediately and not wait until the update / reboot takes place.

If the user wants to have this that the server restarts immediately he could just set WARN_MINUTES to 0 (not sure if that works as expected, haven't tested that).

Dashboy1998: We should still be checking in backup.sh, restore.sh, and update.sh as those are also intended to be started by the user and not just cron.

I would say for another PR we add validation for all variables which are an expected type such as int, float, boolean.

The implementation in 0.26.0 is no validation for AUTO_UPDATE_WARN_MINUTES.

Dashboy1998: Thinking about it some more auto update and auto reboot should be their own script and then have a separate scripts which are to be ran by the user.

Because in my mind the user should be able to run update and tell it how long to wait or reboot instantly whereas now you can run it but you don't set the values.

This should definitely be a different PR.

scripts/auto_reboot.sh
Luatan: Wouldn't it be better, if we added arg flags to the script instead of doing it interactively? Or both. Then we can call the script with the flags from the cronjob and if the user does not supply any flags it asks for them. That way we only have to manage one file.
Just an idea

scripts/update.sh
Dashboy1998: I was thinking the user interactive script would just be a wrapper for the non interactive script.

scripts/helper_functions.sh
```diff
@@ -93,47 +96,205 @@ export YellowBoldText='\033[1;33m'      # Yellow
...
+broadcast_command() {
+    local return_val=0
+    # Replaces spaces with underscore
+    local message="${1// /_}"
+    if [[ $TEXT = *[![:ascii:]]* ]]; then
+        LogWarn "Broadcast message contains non-ascii characters: \"${message}\""
+    fi
+    if ! RCON "broadcast ${message}"; then
+        return_val=1
+    fi
```
Dashboy1998, line 239: @Luatan do you think we should always check if the message being broadcast contains non-ascii characters or only when the message is broadcast?
```suggestion
    if RCON "broadcast ${message}"; then
        # If message is broadcast then log if it contains non-ascii characters
        if [[ $TEXT = *[![:ascii:]]* ]]; then
            LogWarn "Broadcast message contains non-ascii characters: \"${message}\""
        fi
    else
        return_val=1
    fi
```

Luatan: I think we should check before we actually send it. I just tested it a bit. It seems like it does broadcast, but it also gives me an error:
```
docker exec -it palworld-server rcon-cli "broadcast 🙌"
Broadcasted: ð�
cli: execute: rcon: read tcp 127.0.0.1:45320->127.0.0.1:25575: i/o timeout
```
Also there won't be anything "useful" broadcasted anyway.

Dashboy1998:
This has turned into a very big PR. I am going to break into smaller ones.

Dashboy1998:
Still need a PR for countdown function however this will be after #411 and #412

PR #404, opened 2024-02-17
Fix Discord Webhook warn levels

Luatan:
## Context 

* fixes that the server only sends messages with info level

## Test instructions

1. Configure a Webhook in the environment variables
2. Start the server
3. Stop the server
4. Verify that there is sent a message with green and one with red

### DIFF
--- a/scripts/helper_functions.sh
+++ b/scripts/helper_functions.sh
@@ -120,7 +120,7 @@ Log() {
 DiscordMessage() {
   local message="$1"
   local level="$2"
-  if [ -n "$level" ]; then
+  if [ -z "$level" ]; then
     level="info"
   fi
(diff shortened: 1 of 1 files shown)


scripts/helper_functions.sh
```diff
@@ -120,7 +120,7 @@ Log() {
 DiscordMessage() {
   local message="$1"
   local level="$2"
-  if [ -n "$level" ]; then
+  if [ -z "$level" ]; then
```
Dashboy1998, line 123: Good catch

Dashboy1998 (approved):

scripts/helper_functions.sh
Luatan: I was testing something with webhooks and thats where I noticed it. Small but crucial line 😊

PR #408, opened 2024-02-18
Target Manifest - Allow game version lock 

Phelms215:
## Context 

* Utilizes a new environment variable TARGET_MANIFEST_ID to lock to a specific manifest based on the steam depot manifest list. [Found here](https://steamdb.info/depot/2394012/manifests/)
* Allows for easy transition with update system and on and off specific manifests 
* Updates ACF file to ensure version detection functions as expected 

## Choices

* Manifest ID is the easiest most future proof way to lock to a version 
* Moved install/update functions to a "helper_install.sh" 

## Test instructions

1. Utilize TARGET_MANIFEST_ID with other update related environment variables  


thijsvanloef (approved):
Works like a charm, I did add a disclaimer stating that downgrading a save is at your own risk. Thanks!

PR #409, opened 2024-02-19
Fix always showing: "Game not at target version"

thijsvanloef:
## Context 

With the recent addition of the option to pin the manifest this small bug made its way into the logs.
This change makes sure that the message only appears when needed

## Choices

Added a check to make sure that the manifest is set before checking if it matches the current version

## Test instructions

1. Do not set TARGET_MANIFEST_ID
2. start the server
3. notice that it only displays the message when applicable

### DIFF
--- a/scripts/helper_install.sh
+++ b/scripts/helper_install.sh
@@ -91,7 +91,7 @@ UpdateRequired() {
     return 0
   fi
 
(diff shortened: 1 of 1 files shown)


thijsvanloef:
@Phelms215 found a small bug :)

PR #410, opened 2024-02-19
Fixed image size

Dashboy1998:
## Context 

* Reduces image size
* Also delete /tmp/dumps to remove this error when starting as nonroot and a user which does not have a UID of 1000 `/tmp/dumps is not owned by us - delete and recreate`

## Choices

Previously we changed the permission of `/home/steam/steamcmd/` which resulted in an image size increase of 300MB. Now making the directory we need and changing it's permission

## Test instructions

1. Test that is runs with the default compose
2. Test as user 1000:1000
3. Test as any other user

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -112,6 +112,7 @@ RUN chmod +x /home/steam/server/*.sh && \
 WORKDIR /home/steam/server
 RUN touch rcon.yaml crontab && \
     mkdir -p /home/steam/Steam/package && \
+    rm -rf /tmp/dumps && \
     chmod o+w rcon.yaml crontab /home/steam/Steam/package && \
     chown steam:steam -R /home/steam
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -112,6 +112,7 @@ RUN chmod +x /home/steam/server/*.sh && \
 WORKDIR /home/steam/server
 RUN touch rcon.yaml crontab && \
     mkdir -p /home/steam/Steam/package && \
+    chown steam:steam /home/steam/Steam/package && \
     rm -rf /tmp/dumps && \
     chmod o+w rcon.yaml crontab /home/steam/Steam/package && \
     chown steam:steam -R /home/steam/server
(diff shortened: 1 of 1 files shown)


thijsvanloef:
Hi @Dashboy1998 Thanks for making this so quick, however I don't see a size difference: 
```
REPOSITORY                            TAG       IMAGE ID       CREATED          SIZE
thijsvanloef/palworld-test            latest    aae0af4139c1   5 minutes ago    781MB
thijsvanloef/palworld-server-docker   dev       5a1bf2134d3a   18 minutes ago   781MB
thijsvanloef/palworld-server-docker   latest    001d8c947cbc   7 days ago       476MB
```

Dockerfile
```diff
@@ -110,9 +111,10 @@ RUN chmod +x /home/steam/server/*.sh && \
 
 WORKDIR /home/steam/server
 RUN touch rcon.yaml crontab && \
-    chmod o+w rcon.yaml crontab && \
-    chown steam:steam -R /home/steam && \
-    chmod -R o+w /home/steam/steamcmd
+    mkdir -p /home/steam/Steam/package && \
+    rm -rf /tmp/dumps && \
+    chmod o+w rcon.yaml crontab /home/steam/Steam/package && \
+    chown steam:steam -R /home/steam
```
thijsvanloef, line 117: Is this line still supposed to be there?

thijsvanloef: (line 117)

Dashboy1998: I recommend it because the files and directories we make in the build process are not owned by steam. Although it is not needed. Just makes it so the following are always owned by steam even when in rootless

```
/home/steam/server
/home/steam/server/start.sh
/home/steam/server/helper_install.sh
/home/steam/server/helper_functions.sh
/home/steam/server/compile-settings.sh
/home/steam/server/files
/home/steam/server/files/PalWorldSettings.ini.template
/home/steam/server/auto_reboot.sh
/home/steam/server/init.sh
/home/steam/server/discord.sh
/home/steam/server/crontab
/home/steam/server/rcon.yaml
```

thijsvanloef: would `chown steam:steam -R /home/steam/server` work better?

Dashboy1998: > would `chown steam:steam -R /home/steam/server` work better?

Yes. 

I'm getting 465.6MB with either one.

thijsvanloef:
@Dashboy1998 Thanks for taking the time to work with me on this one, gonna test and merge tomorrow

PR #436, opened 2024-02-23
Switch to pidof in restore

### LINKED ISSUE #437: Restore error waiting for server to stop
## Describe the bug

Restore never waits for the server process to stop 

## To Reproduce

Restore

## Expected behavior

No issue

## Actual behavior

```
****Shutting Down Server****
Complete Save
Shutdown...
pgrep: pattern that searches for process name longer than 15 characters will result in zero matches
Try `pgrep -f' option to match against the complete command line.
Shutdown Complete
****Starting Restore****
```

Note Shutdown Complete is lying right now.
```
****Shutting Down Server****
Complete Save
Shutdown...
pgrep: pattern that searches for process name longer than 15 characters will result in zero matches
Try `pgrep -f' option to match against the complete command line.
Still running
Shutdown Complete
****Starting Restore****
```

## Additional context

If you check if the server is still running by modifying restore before the shutdown message can see it's still running
```
server_pid=$(pidof PalServer-Linux-Test)
if [ -n $server_pid ]; then
    echo "Still running"
fi
```

Issue is caused by pgrep. Two solutions exists
1. pgrep `PalServer-Linux` instead of `PalServer-Linux-Test`
2. Switch to pidof and use the whole name (Also eliminates the loop)

Dashboy1998:
## Context 

* Resolves #437 
* Remove extra loop in restore

## Choices

* Switching to pidof since we use pidof in start.sh

## Test instructions

1. Verify restore still works

### DIFF
--- a/scripts/restore.sh
+++ b/scripts/restore.sh
@@ -72,7 +72,7 @@ if [ -f "$BACKUP_FILE" ]; then
         server_pid=$(pidof PalServer-Linux-Test)
         if [ -n "${server_pid}" ]; then
             LogInfo "Waiting for Palworld to exit.."
-            tail --pid="${server_pid}" -f 2>/dev/null
+            tail --pid="${server_pid}" -f /dev/null
         fi
           LogSuccess "Shutdown Complete"
 
--- a/scripts/restore.sh
+++ b/scripts/restore.sh
@@ -74,7 +74,7 @@ if [ -f "$BACKUP_FILE" ]; then
             LogInfo "Waiting for Palworld to exit.."
             tail --pid="${server_pid}" -f /dev/null
         fi
-          LogSuccess "Shutdown Complete"
+        LogSuccess "Shutdown Complete"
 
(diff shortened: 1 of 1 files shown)


Dashboy1998:
Server is currently not stopping during restore fix.

scripts/restore.sh
```diff
@@ -72,7 +72,7 @@ if [ -f "$BACKUP_FILE" ]; then
         server_pid=$(pidof PalServer-Linux-Test)
         if [ -n "${server_pid}" ]; then
             LogInfo "Waiting for Palworld to exit.."
-            tail --pid="${server_pid}" -f 2>/dev/null
+            tail --pid="${server_pid}" -f /dev/null
```
Dashboy1998: Working now.

Not entirely sure why this change was needed as it works in start.sh but not in here.

thijsvanloef (approved):
Tested and restores are working again

PR #453, opened 2024-02-26
Check RCON enabled before detecting port

justforlxz:
## Context 

Before detecting that the RCON port is available, we should check if RCON is selected to be enabled

## Choices

If RCON_ENABLED is configured in the environment variable and RCON_PORT is configured, you need to use nc to check.

## Test instructions

1. 

### DIFF
--- a/scripts/player_logging.sh
+++ b/scripts/player_logging.sh
@@ -13,10 +13,12 @@ get_playername(){
 }
 
 # Wait until rcon port is open
-while ! nc -z 127.0.0.1 "${RCON_PORT}"; do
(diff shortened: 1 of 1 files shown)


thijsvanloef:
I do not think this is necessary, we check if RCON is enabled before starting the player_logging.sh script:
https://github.com/thijsvanloef/palworld-server-docker/blob/668fd61671578cfc6c28fc1f4f1fcc156bfe2d41/scripts/start.sh#L146

justforlxz:
Oh, sorry, I just noticed.

PR #459, opened 2024-02-27
Include RCON Port CLI flag when enabled

bostrt:
## Context 

* In Palworld 1.5.0 update it seems that the `-rconport` flag is required to actually enable RCON

## Test instructions

1. Build and launch container using this PR
2. Run container with default env var settings (rcon enabled and 25575/tcp)
3. Try to connect using `rcon-cli` and run `Info` successfully 

### DIFF
--- a/scripts/start.sh
+++ b/scripts/start.sh
@@ -81,6 +81,10 @@ if [ "${MULTITHREADING,,}" = true ]; then
     STARTCOMMAND+=("-useperfthreads" "-NoAsyncLoadingThread" "-UseMultithreadForDS")
 fi
 
+if [ "${RCON_ENABLED,,}" = true ]; then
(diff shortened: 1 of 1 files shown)


scripts/start.sh
```diff
@@ -81,6 +81,10 @@ if [ "${MULTITHREADING,,}" = true ]; then
     STARTCOMMAND+=("-useperfthreads" "-NoAsyncLoadingThread" "-UseMultithreadForDS")
 fi
 
+if [ "${RCON_ENABLED,,}" = true ]; then
+    STARTCOMMAND+=("-rconport=${RCON_PORT}")
```
MusclePr, line 85: `    STARTCOMMAND+=("-rcon")`
No need to specify port.

Issue #458, opened 2024-02-27
Update for v0.1.5.0


MusclePr:
This is not bug.
The rcon port no longer listen from v0.1.5.0 without `-rcon` or `-rconport=` option.

and modified options.
`EpicApp=PalServer`  ---> `-publiclobby`
reference from https://tech.palworldgame.com/settings-and-operation/arguments

```diff
diff --git a/scripts/start.sh b/scripts/start.sh
index 4e014c9..1b826e2 100644
--- a/scripts/start.sh
+++ b/scripts/start.sh
@@ -65,6 +65,7 @@ isReadable "${STARTCOMMAND[0]}" || exit
 isExecutable "${STARTCOMMAND[0]}" || exit
 
 # Prepare Arguments
+STARTCOMMAND+=("-rcon")
 if [ -n "${PORT}" ]; then
     STARTCOMMAND+=("-port=${PORT}")
 fi
@@ -74,7 +75,7 @@ if [ -n "${QUERY_PORT}" ]; then
 fi
 
 if [ "${COMMUNITY,,}" = true ]; then
-    STARTCOMMAND+=("EpicApp=PalServer")
+    STARTCOMMAND+=("-publiclobby")
 fi
 
 if [ "${MULTITHREADING,,}" = true ]; then
```

urbenlegend:
Hmmm, I wonder if this is probably why I am still experiencing symptoms from this closed bug: https://github.com/thijsvanloef/palworld-server-docker/issues/449#issuecomment-1965751390

dualznz:
yea im getting the constant logging from rcon port not being open when it is

bostrt:
I've submitted a PR for the RCON issue: https://github.com/thijsvanloef/palworld-server-docker/pull/459

dualznz:
yup i implemented that into my start.sh file but i was still getting the flag error, so i commented out the `Waiting for RCON port to open to show player logging...` line in `/home/steam/server/player_logging.sh` then it wont come up all the time, it appears 0.1.5 broke alot of things.

also noticed that when logged into in-game rcon, you no longer see the connected player in the `escape` menu

MusclePr:
add `bShowPlayerList=True` in `PalWorldSettings.ini` from v0.1.5.0

chowhi123:
@43414d  We also need to add new variable bShowPlayerList=True to DefaultPalWorldSettings.ini. I don't know how to make changes using git. Could you help me?

[...]

PR #461, opened 2024-02-27
add default bShowPlayerList=True to PalWorldSettings.ini

MusclePr:
## Context 

* add `bShowPlayerList` part of #458

## Choices

* I think True is better than the default bShowPlayerList=False.
  It will be displayed even if you are not an admin.

## Test instructions

1. press ESC int the world.


chowhi123:
.env.example also requires this change. Thank you!

thijsvanloef:
Please also add documentation to the Docusaurus and readme

thijsvanloef (approved):
works like a charm

PR #473, opened 2024-02-29
Remove -rcon arg statement

43414d:
-rcon is no longer needed as of the 1.5.1 fix, seems like it was bugged in 1.5.0 and not pulling correctly from config

## Context 

Removal of excess statement from 1.5.0 bug

## Choices

It's just boilerplate now since official fix

## Test instructions

1. Reverted back to v0.28 image that didn't have -rcon arg
2. Recomposed server
3. Server rcon still working correctly (player list check)


43414d:
> I'd recommend that you take this one.
> 
> ```shell
> # fix bug for v0.1.5.0 only.
> if [ "${TARGET_MANIFEST_ID}" = "3750364703337203431" -a "${RCON_ENABLED,,}" = true ]; then
>     STARTCOMMAND+=("-rcon")
> fi
> ```

I don't disagree, but with the container automatically updating the game from steam is it viable to put in a legacy fix?

Ie. I can't think of a situation currently where you'd run palworld-server-docker:latest and not have palworld server updated as well.

sonroyaalmerol:
Don't we have the feature to trick steamcmd into downloading an older version of the server with the `TARGET_MANIFEST_ID` env var? If so, then I think it's still a good idea to support legacy fixes unless it compromises the latest versions.

MusclePr:
The reason I recommended it is because of this feature.
https://github.com/thijsvanloef/palworld-server-docker?tab=readme-ov-file#locking-specific-game-version

MusclePr:
However, I think v0.1.5.0 has almost no use other than analyzing save data.

43414d:
Ahh, see I had no idea about that feature. We love productive PR comment conversations.

I don't mind adding it, but yeah we're basically saying "okay sure, if you wanna rollback to a buggy game version, that's all you, bud." If one person finds value in a change then there's bound to be others as well. :)

43414d:
@MusclePr I'm a little unfamiliar with '-a' syntax, does it short circuit or eval both expressions?

MusclePr:
That's a great point!
`-a` seems to be deprecated in POSIX, so I changed the format.
```shell
# fix bug for v0.1.5.0 only.
if [ "${TARGET_MANIFEST_ID}" = "3750364703337203431" ] && [ "${RCON_ENABLED,,}" = true ]; then
    STARTCOMMAND+=("-rcon")
fi
```

43414d:
All good! I had already coded it for && just was checking with you first.

43414d:
definitely wrong commit lol

thijsvanloef (approved):

Issue #324, opened 2024-02-08
Flexible discord notification


Torotin:
Greetings.
I suggest introducing separate variables for Discord messages about working with backups.

DISCORD_BACKUP={true|false}
DISCORD_BACKUP_WEBHOOK_URL="{STRING}"
DISCORD_PRE_BACKUP_MSG="{STRING}"
DISCORD_BACKUP_OK_MSG="{STRING}"
DISCORD_BACKUP_ERROR_MSG="{STRING}"
DISCORD_PRE_BACKUP_DELETE_MSG="{STRING}"
DISCORD_BACKUP_OK_DELETE_MSG="{STRING}"
DISCORD_BACKUP_BACKUP_ERROR_DELETE_MSG="{STRING}"

This change will allow for more flexible notification flows: for users (players) - only server status, and for admins - variable enabling/disabling of notifications with the possibility of directing them to a separate admin channel.

jiannystein:
Just to share and gather any additional insights on how others are doing their things on this topic

My Docker webhook is configured to notify the admin's channel. Here's what it looks like at the moment for backup:
![image](https://github.com/thijsvanloef/palworld-server-docker/assets/41028418/b4bc6ab8-fa95-4ff3-8f2d-09d785832b71)

Script for notifying the public channel and broadcasting via RCON, as well as restarting the server when memory usage exceeds 90%, to prevent unexpected interruptions:
![image](https://github.com/thijsvanloef/palworld-server-docker/assets/41028418/9dae2415-19f8-4b65-9550-3a445a11739d)

Script for notifying the public join/leave channel and broadcasting via RCON when a player joins or leaves:
![image](https://github.com/thijsvanloef/palworld-server-docker/assets/41028418/b1ace498-bd68-4e95-8c66-4273597bed9f)

thijsvanloef:
@Torotin @jiannystein I'm currently working on this enhancement, if you could, could you test this image? `thijsvanloef/palworld-server-docker:pr-490`
| Variable                                   | Info                                                                                                                                                                                                | Default Values                                                                                         | Allowed Values                                                                                                    |
|--------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|--------------------------------------------------------------------------------------------------------|-------------------------------------------------------------------------------------------------------------------|
[...]

PR #487, opened 2024-03-03
Add Discord player logging to notify when players join and leave the server

xHyperElectric:
## Context 

Add discord player logging to notify via Discord webhook when players join and leave the server.

Adds this feature mentioned in #324 

> Script for notifying the public join/leave channel and broadcasting via RCON when a player joins or leaves: image

## Choices

I only 1 line for joining and one line for leaving inside the pre-existing player_logging.sh loops that are currently used for broadcasting to the server. I added lines to call the pre-existing function DiscordMessage() to send a message in the loop.

I added two env variables to allow the ability to change the join and leave messages. (DISCORD_PLAYER_LEAVE_MESSAGE and DISCORD_PLAYER_JOIN_MESSAGE)

I'm quite new to docker so I could be wrong, but I believe the way I did it would work. In the dockerfile I left ${player_name} in both messages for them to be substituted later. If I understand correctly, the string "${player_name}" can not be removed if the message is changed, otherwise the substitution would fail.

I did not add anything in the message to indicate how long a player was on the server before they left.

## Test instructions

I am unable to test it myself, therefore I haven't.

1.) Create a DISCORD_WEBHOOK_URL
2.) Have a player join or leave the server
3.) Check for a message in the channel for which the DISCORD_WEBHOOK_URL is set up to post to


xHyperElectric:
I'm not familiar with docusaurus but it seems the check failed because I didn't properly escape the ${player_name} in the docusaurus docs.

```Error: Docusaurus server-side rendering could not render static page with path /getting-started/configuration/server-settings because of error: player_name is not defined```

PR #491, opened 2024-03-06
Refactor Function names to snake case

thijsvanloef:
## Context 

Since the function names are all over the place, from now on, only snake_case will be supported.
Refactoring the helper_functions and helper install to only include snake_case functions

## Choices

https://google.github.io/styleguide/shellguide.html#s7.1-function-names

## Test instructions

1. 


thijsvanloef:
Decided to do this in steps, not all at once

PR #540, opened 2024-04-24
Modifying Supercronic Related Code in start.sh

kimzuni:
## Context 

1. Before registering Supercronic, check the size of the `crontab` file, not variables value.
2. If you perform the task registered in Supercronic, the log as like:
image

## Choices

1. If you modify it this way, you don't need to modify it even if related functions are added.
2. Add the `-passthrough-logs` option to Supercronic so that the standard output of the file is displayed normally.
image

## Test instructions

1. Set the `BACKUP_ENABLED` variable to `false` so that nothing is stored in the `crontab` file, and check whether the `else` part is executed.
2. Set the `BACKUP_ENABLED` or `AUTO_UPDATE_ENABLED` variable to `true` to see if it goes to the `if` part, and check the log on which `update` was executed.

### DIFF
--- a/scripts/start.sh
+++ b/scripts/start.sh
@@ -153,9 +153,8 @@ if [ "${AUTO_REBOOT_ENABLED,,}" = true ] && [ "${RCON_ENABLED,,}" = true ]; then
     supercronic -quiet -test "/home/steam/server/crontab" || exit
 fi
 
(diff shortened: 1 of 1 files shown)


thijsvanloef:
Love this change!

Issue #640, opened 2024-12-17
Orange Pi 5 Max Performance improvement


DummyTitan:
## Is your feature request related to a problem? Please describe

The main issue is that at first it was running perfectly, with of course a low fps (23 to 25 fps). The thing is, the Orange Pi 5 processor is double the performance of the RPi5, and I know of people running this fine in their RPi5. Perhaps it requires a little bit of coding or research, but no one has any info on the net currently. 

Whenever I was playing with my cousin, the system tended to lag a little, but it wasn't much. After a while playing and starting our bases, a lot of lag started happening, and traspasing floors or walls wasn't uncommon, which is frustrating, and I believe it might have something to do with the emulation or some kind of setting I put.

## Describe the solution you'd like

As such, if there is any way that you could add to the ARM64_DEVICE some kind of setting for the rk3588 processor, which includes multiple SOC, it would be great; if it cannot be done, no problem. And if you could make a template for making it run in something like Puferpanel, which would allow me to see in live the issues presented in the Docker, it might also help (tried with pterodactyl and it had many issues with the panel so i was unable to install it).

## Describe alternatives you've considered

I have attempted turning off the multithreading and turning it on, also changing the settings for the "BOX64_DYNAREC_STRONGMEM",
    "BOX64_DYNAREC_BIGBLOCK", "BOX64_DYNAREC_SAFEFLAGS", "BOX64_DYNAREC_FASTROUND", "BOX64_DYNAREC_FASTNAN",
"BOX64_DYNAREC_X87DOUBLE".

## Additional context

[env.txt](https://github.com/user-attachments/files/18171578/env.txt)
[json-code.txt](https://github.com/user-attachments/files/18171581/json-code.txt)
The env.text file is the .env file I used after changing the settings a little to see if there was any improvement, and the json-code.txt is a code I made to use as a template in pufferpanel that, unfortunately, didn't get farther than pulling the Docker image.

## Feature Report Checklist

- [ ] I am willing to implement this feature

sonroyaalmerol:
I've added `ARM64_DEVICE=rk3588` in the arm64 base image. You'll be able to test it as soon as @thijsvanloef updates the image tag. I'm not familiar with Puferpanel (and it's out of the "arm64-exclusive" scope) so I'll leave that to him as well :)

thijsvanloef:
[...]

PR #664, opened 2025-01-19
Refactor InstallServer function to integrate DepotDownloader better

sonroyaalmerol:
## Context 

This is mostly to help #660 by updating DepotDownloader to the latest version and reorganize the installation script. I couldn't recreate the issue on my own ARM64 machine so I'm not sure if this would solve it. I'm hoping it has something to do with this change in the latest version:
> Do not abort downloading if one depot fails to get depot key

At this point, DepotDownloader is our only option for Apple CPUs (and probably other ARM64 processors) to download the server files. Refactoring the script to integrate it with steamcmd even if it's not explicitly enabled will probably be a net positive for the user experience.

## Choices

- Force ARM64 hosts without 4k page size to use DepotDownloader because steamcmd will never work on them anyways.
- Always fallback to DepotDownloader for all steamcmd commands that fail (non-exit code 0). I don't think there's any disadvantage to doing this at all.
- Check PalServer.sh existence before creation of PalServer-arm64.sh to prevent "non user-friendly" errors returned by the cp, sed, and chmod commands. This is just purely for better experience and log flow.
- I also added `-NumberOfWorkerThreadsServer=$(nproc --all)` while I'm at it as suggested in #640 since it's a pretty small change for an entirely new PR.

## Test instructions

1. Ran the container in both architectures without the /palworld folder
2. Ran the container in both architectures with the /palworld folder

- [ ] I've added documentation about this change to the [docs](https://github.com/thijsvanloef/palworld-server-docker/tree/main/docusaurus/docs).


sonroyaalmerol:
This should be ready to go. The amd64 unit test seems to need more timeout as it's taking longer for it to download the server for some reason.

Issue #702, opened 2025-06-25
Palworld Terraria update thread


thijsvanloef:
Hi all!

The Terraria update dropped.

If you encounter any issues with the Docker container, please report them by commenting on this thread. Include detailed steps to reproduce the issue, relevant logs, and any additional context that could help in diagnosing the problem.

Thanks in advance,

Thijs

deafire:
error:
Encountered unexpected error downloading chunk 2ad255c5c51318be49ef7b27c2bebb4bc14b9778: End of Central Directory record could not be found.

[message.txt](https://github.com/user-attachments/files/20896736/message.txt)

deafire:
> > 错误：下载区块 2ad255c5c51318be49ef7b27c2bebb4bc14b9778 时遇到意外错误：找不到中央目录记录的结尾。
> > [message.txt](https://github.com/user-attachments/files/20896736/message.txt)
> 
> 请检查以下步骤是否解决了您的问题（BACKUP EVERYTHING BEFORE）：
> 
> ```
> docker compose down --rmi all
> ```
> 
> * 删除以下文件：./palworld/steamapps/appmanifest_2394010.acf
> 
> ```
> docker compose up -d
> ```

cannot be resolved

MusclePr:
@deafire 
I had the same problem, so I submitted a pull request to fix it.
Build #703 locally or wait for it to be merged.

build & overwrite image:
```bash
git clone https://github.com/MusclePr/palworld-server-docker.git
cd palworld-server-docker
git switch fix/certificate-has-expired
docker build -t thijsvanloef/palworld-server-docker:latest .
```

thijsvanloef:
This might be related to depotdownloader being out of date, testing it currently. Please be patient for a bit longer while I try to figure it out 😃

thijsvanloef:
Release 1.4.1 has been made, `:latest` will be updated in about 10 min 😄

oALeixo:
Hi, I'm facing exact same error

UpdateUI: skip show logo

[----] !!! Fatal Error: Steamcmd needs to be online to update. Please confirm your network connection and try again.

src/tier0/threadtools.cpp (3601) : Assertion Failed: Illegal termination of worker thread 'Thread(0x0x5c57ee30/0x0xeddafb'

SteamCMD failed, falling back to DepotDownloader

No username given. Using anonymous account with dedicated server subscription.

Connecting to Steam3...Connection to Steam failed. Trying again (#1)...

[...]

PR #703, opened 2025-06-25
update ca-certificates

MusclePr:
## Context 

Some ca certificates has expired.
Fix download failure.

## Choices

update ca-certificates.

## Test instructions

1. Verify that the Terraria update fails.
2. Apply this PR.
3. Check if the update was successful.

- [ ] ~~I've added documentation about this change to the [docs](https://github.com/thijsvanloef/palworld-server-docker/tree/main/docusaurus/docs).~~


MusclePr:
Maybe, Fixes issue #702

thijsvanloef:
https://github.com/thijsvanloef/palworld-server-docker/pull/704 This seems to have fixed the depotdownloader issue, is this PR still relevant?

MusclePr:
If the problem is resolved, this PR is unnecessary.

Updating the certificate is not required, but it is a good idea to update it.

In my environment, curl was failing due to an expired certificate, but this problem can no longer be reproduced.


```log
****Starting Installation****
Downloading server files with SteamCMD
Redirecting stderr to '/home/steam/Steam/logs/stderr.txt'
Logging directory: '/home/steam/Steam/logs'
[  0%] Checking for available updates...
curl: (60) SSL certificate problem: certificate has expired
More details here: https://curl.se/docs/sslcerts.html

curl failed to verify the legitimacy of the server and therefore could not
establish a secure connection to it. To learn more about this situation and
how to fix it, please visit the web page mentioned above.
```

I don't think this problem will occur unless you select a download server with an expired certificate.

I recommend installing it just in case, as curl and wget are used other than DepotDownloader as well.

thijsvanloef:
Agreed

Issue #846, opened 2026-06-18
Fresh install fails to run due to execute permissions


Araeos:
## Describe the bug

When the docker image is run and after downloading the first files using the fallback DepotDownloader (since the SteamCMD fails due to  "needs to be online to update") the startup of fails silently. The palworld volume contains `PalServer.sh` , among others, but this file is missing the executable permission. After fixing this permission, the server continues to boot successfully.

## To Reproduce

1. Delete the contents of the palworld volume 
2. Start the palworld server `docker compose up`
3. Wait until download is finished
4. The docker container closes automatically, stopping without giving an error (and restarts if configured to do so)
5. (fix) Execute `chmod +x PalServer.sh` to fix permissions manually

## Expected behavior

The palworld server should run in the container and create a new save folder.

## Actual behavior

The docker run stops and may restart if configured to do so.

## OS information (IMPORTANT)

- OS:  Up-to-date headless arch linux
- Palworld Version:  0.7.3.90464
- Docker Image Version:  thijsvanloef/palworld-server-docker:latest

## Hardware information (IMPORTANT)

- CPU:  AMD Ryzen 7 5700G
- Memory:  32 GB

## compose.yaml contents (IMPORTANT)

<details><summary>compose.yaml</summary>

```yaml
services:
   palworld:
      image: thijsvanloef/palworld-server-docker:latest
      restart: no
      container_name: palworld-server
      stop_grace_period: 60s # Set to however long you are willing to wait for the container to gracefully stop
      ports:
        - 8211:8211/udp
        - 27015:27015/udp
        # - 8212:8212/tcp  # Port for REST API if REST_API_ENABLED: true
      env_file:
         -  /etc/palworld/environment.env
      volumes:
         - /var/lib/palworld:/palworld
```

</details>

## Container/Host Logs (IMPORTANT)

<details><summary>Logs</summary>

```log
[1;36m****EXECUTING USERMOD****[0m
[1;36m****AUTO PAUSE with Community****[0m
[0;37mLaunch proxy.[0m
[0;37mWait until proxy is initialized...[0m
[0;37mProxy initialized.[0m
[0;37mUsing proxy now.[0m
[0;37mServer installation not detected.[0m
[1;36m****Starting Installation****[0m
[1;33mDownloading server files with SteamCMD[0m
[0;37mWarming up SteamCMD...[0m
Redirecting stderr to '/home/steam/Steam/logs/stderr.txt'
[...]

PR #847, opened 2026-06-19
Added logging to failed checks in start.sh

Dashboy1998:
## Context 

Fixing #846 

## Choices

Checking two types of permissions before exiting so the user gets both messages instead of fixing the first then running again only to get the second one.

## Test instructions

Disable container restart
Run it with incorrect permissions
Start the container after each step.

Note for step 2: Directories may be writable but not deletable causing -w to return false so for isWritable we check if a file is able to be made but since step 2 permission aren't executable you cannot run mktemp so it counts it as not writable. Keeping it this way as without it we would have the issue presented in #358 

|  # | Action Outside Container | Expected Output |
|--------------------|--------------------------------|------------------------------------|
|  1 | `chmod ugo=r palworld/` | /palworld/ is not writable. /palworld/ is not executable. |
|  2 | `chmod ugo=rw palworld/` | /palworld/ is not executable. /palworld/ is not executable.|
|  3 | `chmod ugo=rx palworld/` | /palworld/ is not writable. |
|  4 | `chmod ugo=rwx palworld/` | Server Starts |
|  5 |  `chmod ugo=rwx palworld/PalServer.sh` | Server Starts |
|  6 | `chmod ugo=w palworld/PalServer.sh` | ./PalServer.sh is not readable. ./PalServer.sh is not executable.|
|  7 | `chmod ugo=r palworld/PalServer.sh` | ./PalServer.sh is not executable. |
|  8 | `chmod ugo=x palworld/PalServer.sh` | ./PalServer.sh is not readable. |

### DIFF
--- a/scripts/helper_functions.sh
+++ b/scripts/helper_functions.sh
@@ -8,7 +8,7 @@ dirExists() {
     local path="$1"
     local return_val=0
     if ! [ -d "${path}" ]; then
-        echo "${path} does not exist."
+        LogError "${path} does not exist."
         return_val=1
     fi
     return "$return_val"
@@ -21,7 +21,7 @@ fileExists() {
     local path="$1"
     local return_val=0
     if ! [ -f "${path}" ]; then
-        echo "${path} does not exist."
+        LogError "${path} does not exist."
         return_val=1
     fi
     return "$return_val"
@@ -34,7 +34,7 @@ isReadable() {
     local path="$1"
     local return_val=0
     if ! [ -e "${path}" ]; then
-        echo "${path} is not readable."
+        LogError "${path} is not readable."
         return_val=1
     fi
     return "$return_val"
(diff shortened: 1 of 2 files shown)


scripts/helper_functions.sh
```diff
@@ -33,8 +33,8 @@ fileExists() {
 isReadable() {
     local path="$1"
     local return_val=0
-    if ! [ -e "${path}" ]; then
```
Dashboy1998, line 36: -e FILE - True if the FILE exists and is a file, regardless of type (node, directory, socket, etc.).
-r FILE - True if the FILE exists and is readable.

scripts/start.sh
thijsvanloef, line 1: The echo to LogError changes look good, however the change in the start.sh script I'm not a big fan of, it seems like a messy way of doing it.

Dashboy1998: I'm not a big fan of it either but we could just go back to it exiting instead of checking the next permission.

Dashboy1998: We can keep it the current way. Maybe I'll make an issue for it to do both but for now let's just fix one issue.

PR #877, opened 2026-07-11
ARM fix

jiwidi:
Trying to work on arm fix


thijsvanloef:
@jiwidi Are you still working on this?

jiwidi:
@thijsvanloef Hi! No, i'm running this image now, seems more stable with these flags but it also autorestarts when crashing due to new error catching (didnt before). 

Is "okay" for my friends group

lhw:
Has FeX been investigated as an alternative solution to box64? With Valves support it might be a better option stability/performance wise. I might work on that after some other things. I wanted to check how well it works for amd64 servers anyhow.

thijsvanloef:
@lhw I've seen another image that used it, and was looking into it, but like a lot of interesting stuff, was not able to get around to it yet.

PR #884, opened 2026-07-14
feat: allow overriding worker thread count via WORKER_THREADS

oxymoron0:
## Problem

When `MULTITHREADING=true`, the start command hardcodes the worker thread count:

```bash
STARTCOMMAND+=("-useperfthreads" "-NoAsyncLoadingThread" "-UseMultithreadForDS" "-NumberOfWorkerThreadsServer=$(nproc --all)")
```

`nproc --all` reports **all host cores and ignores cgroup CPU limits** (and even CPU affinity). Under Kubernetes or `docker run --cpus=N`, this spawns far more worker threads than the CPU quota allows — e.g. 12 threads under a 4-CPU limit on a 12-core node. The thread burst exhausts the CFS quota early in each 100ms period, freezing the whole server process for the remainder of the period. Players experience this as periodic rubberbanding.

## Change

Add an optional `WORKER_THREADS` environment variable to override the count:

```bash
"-NumberOfWorkerThreadsServer=${WORKER_THREADS:-$(nproc --all)}"
```

- **Behavior is unchanged when the variable is not set** (defaults to the current `$(nproc --all)`).
- On Kubernetes it composes nicely with the Downward API, so the value tracks the CPU limit automatically:

```yaml
env:
  - name: WORKER_THREADS
    valueFrom:
      resourceFieldRef:
        resource: limits.cpu
```

Also adds a `WORKER_THREADS` row to the README environment variable table.

## Verification (production, Kubernetes)

4-CPU limit on a 12-core node, `MULTITHREADING=true`:

| Condition | threads=12 (before) | threads=4 (after) |
|---|---|---|
| Idle, 3h | 20 CFS throttle events / 2.5s throttled | 0 / 0ms |
| Multi-player session, 45+ min | player-reported rubberbanding | 0 / 0ms, no rubberbanding |

(Throttling measured via cgroup v2 `cpu.stat` `nr_throttled`/`throttled_usec`.)


thijsvanloef:
Thank you for the PR, but i've just merged a change that already addresses this 😓 
github.com/thijsvanloef/palworld-server-docker/pull/881

Issue #888, opened 2026-07-16
[ARM64/M4/macOS] Palworld 1.0 crashes with "Detected negative delta time" despite full recommended Box64 stability settings


linuxfreebird:
# [ARM64 / Apple M4 / macOS] Palworld 1.0 crashes with “Detected negative delta time” under Box64

## Describe the bug

The Palworld dedicated server starts successfully on an Apple M4 Mac mini using the ARM64 Docker image, accepts Steam players, loads the world, and allows normal gameplay. After an unpredictable amount of time, the Palworld process crashes with:

```text
LowLevelFatalError
Detected negative delta time
Segmentation fault
```

We followed this guide (https://palworld-server-docker.loef.dev/known-issues). The container then restarts automatically, and connected players are kicked. The same crash still occurs after applying the complete ARM64 stability configuration recommended on the project’s Known Issues page:

```yaml
ARM64_DEVICE: m1
BOX64_DYNAREC_BIGBLOCK: 0
BOX64_DYNAREC_SAFEFLAGS: 2
BOX64_DYNAREC_STRONGMEM: 3
BOX64_DYNAREC_FASTROUND: 0
BOX64_DYNAREC_FASTNAN: 0
BOX64_DYNAREC_X87DOUBLE: 1
```

A separate crash was also previously observed with:

```text
free(): invalid next size (normal)
Aborted
```

This does not appear to be an out-of-memory event. During one measurement, the container used about 2.327 GiB of a 9.703 GiB Docker memory limit, and the logs contained no `OOM`, `Killed`, or memory-limit error.

## To Reproduce

1. Install Docker Desktop on an Apple M4 Mac running macOS Tahoe 26.5.2.
2. Configure the container with `USE_DEPOT_DOWNLOADER: true`, `ARM64_DEVICE: m1`, and all six conservative Box64 settings shown below.
3. Run `docker compose up -d`.
4. Wait until the logs show `Running Palworld dedicated server on :8211`.
5. Join from a Steam Palworld client.
6. Play normally or leave the server running.
7. After an unpredictable amount of time, the server reports `Detected negative delta time`, then segfaults and restarts.

The issue is intermittent. It has happened after about 1 hour 37 minutes, after about 24 minutes, shortly after an automatic restart, with one player, and during sessions with multiple players.

## Expected behavior

The dedicated server should remain running continuously and accept players without Unreal Engine detecting that time moved backward.

## Actual behavior

[...]

PR #892, opened 2026-07-18
Add opt-in mitigation for Palworld negative DeltaTime crashes

Lyinsun:
## Context

Related to #888.

For Palworld `v1.0.1.100619`, binary inspection shows that the server registers the built-in console variable `Pal.AllowNegativeDeltaTime` with a default of `0`. When enabled, its negative-delta branch substitutes approximately `10 ms` instead of entering the fatal path. Runtime inspection separately confirmed that an Engine ini command-line override changes the active CVar state.

This PR exposes that existing Palworld behavior through a default-off environment variable:

```dotenv
PALWORLD_ALLOW_NEGATIVE_DELTA_TIME=false
```

When explicitly enabled, the entrypoint appends exactly one fixed argument to the PalServer command:

```text
-ini:Engine:[ConsoleVariables]:Pal.AllowNegativeDeltaTime=1
```

Unset or `false` preserves the current command line. Empty or otherwise invalid values fail before installation or update work begins.

This is a game-level mitigation for the negative-DeltaTime termination. It does not correct or synchronize the macOS, Docker Desktop/LinuxKit, VM, or Box64 clocks.

## Choices

- The option is disabled by default, so existing deployments retain their current behavior.
- Only a fixed literal is appended through the existing Bash command array; the environment value cannot supply arbitrary startup arguments.
- The same shared startup path supports AMD64 and the existing ARM64/Box64 wrapper without replacing either launcher.
- The value is validated as a case-insensitive `true` or `false`; misspellings and explicitly empty values fail fast instead of silently disabling the mitigation.
- The setting is documented in `.env.example`, the README, and the Docusaurus server-settings page.

This PR does not add a global preload clock shim. The AMD64/Rosetta shim discussed in #888 remains a separate workaround, and its Box64 route was not validated there. A global shim would also change clock semantics for every dynamically linked process in the container, whereas this option only enables Palworld's own recovery branch.

Separate `free()` or `realloc()` heap-corruption aborts are outside this change; one such startup abort was observed independently without a negative-DeltaTime event.

## Validation

The PR adds:

- Architecture-independent helper tests for unset, `false`, case-insensitive `true`, invalid, and explicitly empty values.
- AMD64 container checks confirming invalid and empty values exit with status `1`.
- An AMD64 process check confirming the argument is absent by default.
- An ARM64 process check confirming the enabled argument appears exactly once.

After these focused assertions, the existing jobs continue with their pre-existing normal server checks.

During earlier field validation of the same functional implementation, a concurrent same-host observation captured the unmodified image reporting `DeltaTime=-0.002185`, segfaulting, and restarting. The opt-in candidate remained healthy with restart count `0` and completed a REST save afterward.

A roughly 90-second real-client server-side smoke test confirmed join, movement activity, continued online save, `52-55 FPS`, and no server restart or fatal log during the observation.

A later Docker Desktop backend sampling interval showed a `-111.147 ms` change in its wall-minus-monotonic offset while the opt-in server remained healthy. This is not a measured `-111 ms` PalServer delta: the 30-second sample cannot distinguish a step, slew, or brief stall, or prove propagation into LinuxKit, Box64, or PalServer.

## Test instructions

- `./tests/test-negative-delta-recovery.sh`
- Bash syntax checks for the modified and new scripts
- YAML parse checks for the modified workflows
- `git diff --check` against `upstream/main`
- Independent scope and correctness reviews of the final commit

- [x] I've added documentation about this change to the docs
- [x] I've not introduced breaking changes
- [x] My changes do not violate linting rules


Lyinsun:
Update: the workflows were approved and 10 checks passed. `Docker - Test (arm64)` did not report a code assertion failure; GitHub’s only failure annotation says the self-hosted runner lost communication with the server. The job stopped during `Wait for server to start`, before the new negative-delta process-argument assertion or the remaining integration checks ran.

The PR author cannot rerun this upstream workflow (`Must have admin rights to Repository`). Could a maintainer please rerun the failed job once `oracle-arm64` is healthy? I have not changed the code in response because no code test failed.

Lyinsun:
@thijsvanloef This is now ready for review. Ten checks are green. The only non-green check is `Docker - Test (arm64)`; GitHub annotated it because the self-hosted runner lost communication during `Wait for server to start`, before the new ARM64 process-argument assertion ran. The fork author does not have permission to rerun upstream workflow jobs. Could you please rerun the failed job once the runner is healthy?

Lyinsun:
Update for the new HEAD `e13e3c6771c3c4e7a5482f337a439a75d22e0b7f`:

The maintainer-triggered ARM64 retry reached `oracle-arm64` and built the image, but the existing workflow set `USE_DEPOT_DOWNLOADER=True`. The installer uses a case-sensitive exact comparison with lowercase `true`, so the job actually selected SteamCMD. SteamCMD then returned App 2394010 state `0x202` without installing PalServer, and `Wait for server to start` timed out before the negative-delta assertion ran.

I applied the minimal one-line CI fix (`True` to `true`) in `e13e3c6`. No product code was changed. The exact selector change previously passed the fork ARM64 workflow with DepotDownloader, the negative-delta argv assertion, REST, ports, backup, settings, and stop checks.

The new upstream workflows currently show `action_required` with zero jobs, so they are waiting for repository approval rather than reporting a code failure. @thijsvanloef could you please approve the new workflow runs for this HEAD? I will not change the branch further unless the new CI or review feedback identifies a confirmed issue.

thijsvanloef:
Hey! @Lyinsun it seems like the arm host ran out of space, going to clean it up and restart the job

.github/workflows/linting.yml
```diff
@@ -18,6 +18,9 @@ jobs:
       - name: Shellcheck
         uses: ludeeus/action-shellcheck@2.0.0
 
+      - name: Test negative delta recovery argument
```
thijsvanloef: This test should not be part of the "linting" workflows. A better place would be the unit-test.yml

.github/workflows/unit-test.yml
```diff
@@ -27,6 +27,52 @@ jobs:
           tags: ${{ github.run_id }}
           platforms: linux/amd64
 
+      - name: Reject invalid negative delta recovery setting
```
thijsvanloef, line 33: Really appreciate you adding unit-tests!

scripts/start.sh
```diff
@@ -72,6 +80,8 @@ if [ "${ENABLE_PERF_THREADING_ARGS,,}" = true ]; then
     STARTCOMMAND+=("-useperfthreads" "-NoAsyncLoadingThread" "-UseMultithreadForDS")
 fi
 
+AppendNegativeDeltaRecoveryArgument
```
thijsvanloef: I appreciate the functions but I would like to have it be added to the start command the same way the current appends work.

thijsvanloef (changes requested):

.github/workflows/linting.yml
Lyinsun: You are right — this test should indeed live in unit-test.yml. I moved it there in edc05db.

.github/workflows/unit-test.yml
Lyinsun: Haha, tests are becoming a lifeline in the vibe-coding era. They definitely cannot be taken lightly.

scripts/start.sh
Lyinsun: You are right — this is more consistent with the existing startup argument appends. I updated start.sh to append it directly in edc05db.

thijsvanloef (approved):
LGTM, ty for the contribution!

PR #910, opened 2026-07-28
Handle WorldOption.sav blocking settings on Windows->Linux migration (fixes #886)

### LINKED ISSUE #886: [LOG] REST accessed endpoint / Unauthorized (AdminPassword is empty)
> I always get below error when I enable `ENABLE_PLAYER_LOGGING`, Admin password has been set in env var without any special character. I can also control game server by using that admin password -
> 
> ```
> │ [AUTO PAUSE] Service ... start                                                                                                                        │
> │ [2026-07-15 16:35:21] [LOG] REST accessed endpoint / Unauthorized (AdminPassword is empty)                                                            │
> │ jq: parse error: Invalid numeric literal at line 1, column 13                                                                                         │
> │ [2026-07-15 16:35:51] [LOG] REST accessed endpoint / Unauthorized (AdminPassword is empty)                                                            │
> │ jq: parse error: Invalid numeric literal at line 1, column 13                                                                                         │
> │ [2026-07-15 16:36:21] [LOG] REST accessed endpoint / Unauthorized (AdminPassword is empty)                                                            │
> │ jq: parse error: Invalid numeric literal at line 1, column 13                                                                                         │
> │ [2026-07-15 16:36:51] [LOG] REST accessed endpoint / Unauthorized (AdminPassword is empty)                                                            │
> │ jq: parse error: Invalid numeric literal at line 1, column 13                                                                                         │
> │ [2026-07-15 16:37:21] [LOG] REST accessed endpoint / Unauthorized (AdminPassword is empty)                                                            │
> │ jq: parse error: Invalid numeric literal at line 1, column 13                                                                                         │
> │ [2026-07-15 16:37:51] [LOG] REST accessed endpoint / Unauthorized (AdminPassword is empty)                                                            │
> │ jq: parse error: Invalid numeric literal at line 1, column 13                                                                                         │
> │ [2026-07-15 16:38:21] [LOG] REST accessed endpoint / Unauthorized (AdminPassword is empty)                                                            │
> │ jq: parse error: Invalid numeric literal at line 1, column 13                                                                                         │
> │ [2026-07-15 16:38:51] [LOG] REST accessed endpoint / Unauthorized (AdminPassword is empty)                                                            │
> │ jq: parse error: Invalid numeric literal at line 1, column 13                                                                                         │
> │ [2026-07-15 16:38:51] [LOG] REST accessed endpoint / Unauthorized (AdminPassword is empty)  
> ```
> 
> I found the issue.
> 
> Windows server generates a file called `WorldOption.sav`, this file has a higher priority than `PalWorldSettings.ini`.
> I migrated the game data from Windows to Linux, but the `WorldOption.sav` file prevented `PalWorldSettings.ini` from taking effect.
> 
> As a result, if you have migrated your data from a Windows server, you must delete WorldOption.sav. Otherwise, the Linux server settings will not be successfully applied. 

 _Originally posted by @CloudGee in [#834](https://github.com/thijsvanloef/palworld-server-docker/issues/834#issuecomment-4978875003)_

NathanFant:
## Context

Fixes #886.

A `WorldOption.sav` carried over from a save that was ever played on Windows (including via Steam's "host from save data" co-op mode) takes priority over `PalWorldSettings.ini` and silently prevents the new server's settings -- including `AdminPassword` -- from applying. This causes RCON and the REST API to reject a correctly-configured admin password with "AdminPassword is empty" / `Authentication failed!`, even when that password is verified correct in the container environment and in the generated `PalWorldSettings.ini`.

#886's original report identified the root cause and confirmed deleting `WorldOption.sav` fixes it, but no code change landed -- the existing `migration/` docs and script don't mention this at all, so anyone following the officially documented migration path still hits it.

## Changes

- `migration/migrate.sh`: after copying the save into place, moves aside (does not delete) any pre-existing `WorldOption.sav` in the migrated save directory, so scripted migrations are unaffected by default.
- `migration/README.md`: documents the same step for manual migrations (new step 4), plus a callout at the top of the script instructions explaining why this matters and linking to #886.

Chose "move aside" (`WorldOption.sav` -> `WorldOption.sav.bak`) rather than deleting outright, so the original file is still recoverable if anyone needs it for reasons unrelated to this bug.

## Verification

Reproduced and fixed against a real production migration (Windows-hosted save -> Linux dedicated server, via this same image):

1. Confirmed `AdminPassword` was correct in the container env and in the freshly-generated `PalWorldSettings.ini` (byte-for-byte, matching quoting) -- RCON and REST API (`/v1/api/info`) both still rejected it.
2. Ruled out a config-regeneration race by setting `DISABLE_GENERATE_SETTINGS=true` and restarting against an untouched, already-correct ini -- symptom persisted identically.
3. Ruled out client-side request formatting by manually constructing the Base64 `Authorization: Basic` header and comparing byte-for-byte against `curl -u`'s automatic handling -- identical, and the server's own 401 response confirmed the failure is server-side (`WWW-Authenticate: Basic realm="Pal"`, body `Unauthorized (AdminPassword is empty)`).
4. Moved `WorldOption.sav` aside (this PR's fix) and restarted -- RCON and REST API immediately worked. Confirmed via the server's own logs: `RCON executed the command. Save`, `RCON executed the command. ShowPlayers`. REST API's `/v1/api/info` and `/v1/api/players` returned real, correctly-authenticated responses. All custom settings (drop-item caps, building limits, invader-enemy toggle, etc.) remained correctly applied from `PalWorldSettings.ini` -- this fix does not reset or lose any other configuration.

- [x] I've not introduced breaking changes (the moved-aside file is inert; nothing reads `WorldOption.sav.bak`)
- [x] My changes do not violate linting rules

### DIFF
--- a/migration/README.md
+++ b/migration/README.md
@@ -7,6 +7,16 @@
 >
 > Please make sure you always have a backup!
 
+> [!IMPORTANT]
+> If you're migrating a save that was ever played on Windows (including via Steam's
+> "host from save data" co-op mode), that save's `WorldOption.sav` takes priority over
+> `PalWorldSettings.ini` and will silently prevent your new server's settings —
+> including `AdminPassword` — from applying. This causes RCON/REST API authentication
+> to fail with "AdminPassword is empty" even when the password is correctly configured
+> everywhere else (see #886). `migrate.sh` (below) handles this automatically by moving
+> any pre-existing `WorldOption.sav` aside; if you migrate manually, do the same
+> yourself (see step 3 in "Manually" below).
+
 1. Find a directory which is named by game server name and contains all saved game data,
    usually it will at `~/Steam/steamapps/common/PalServer/Pal/Saved/SaveGames/0/`
 2. Make sure `migration/migrate.sh`, saved game data directory and mounted volume
@@ -44,4 +54,9 @@
    DedicatedServerName=`2E85FD38BAA792EB1D4C09386F3A3CDA`.
 3. Delete the entire new server save at `PalServer\Pal\Saved\SaveGames\0\<your_save_here>`,
    and replace it with the folder from the old server.
-4. Restart the new server
+4. If the copied save folder contains a `WorldOption.sav` file and the save was ever played on
+   Windows, move (don't delete outright — keep it as a backup) that file out of the way, e.g.
+   `mv WorldOption.sav WorldOption.sav.bak`. Otherwise it takes priority over
+   `PalWorldSettings.ini` and silently blocks your new server's settings (including
+   `AdminPassword`) from applying — see #886.
+5. Restart the new server
--- a/migration/README.md
+++ b/migration/README.md
@@ -15,7 +15,7 @@
 > to fail with "AdminPassword is empty" even when the password is correctly configured
 > everywhere else (see #886). `migrate.sh` (below) handles this automatically by moving
 > any pre-existing `WorldOption.sav` aside; if you migrate manually, do the same
-> yourself (see step 3 in "Manually" below).
+> yourself (see step 4 in "Manually" below).
 
 1. Find a directory which is named by game server name and contains all saved game data,
    usually it will at `~/Steam/steamapps/common/PalServer/Pal/Saved/SaveGames/0/`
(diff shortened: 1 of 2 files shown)


NathanFant:
Looks like copilot hung up on linting. I fixed this locally and use it currently. Leaving review and fixing up to y'all.

Issue #906, opened 2026-07-23
Add Linux UE4SS to the image, to allow modding


thijsvanloef:
## Is your feature request related to a problem? Please describe

https://github.com/XarminaEu/ue4ss-linux

## Describe the solution you'd like
Add an ENV var, `ENABLE_UE4SS` to allow modding of the palworld server
You would also need to mount the modding folder if that is not already included in the mounted palworld folder

## Describe alternatives you've considered

Windows UE4SS with Proton/Wine, however not really feasible, as that would practically require a full rewrite to .exe files instead of the used script

## Additional context

<html><body>
<p dir="auto"><strong>This is a Linux build. It CANNOT load Windows <code>.dll</code> files.</strong> Any mod or plugin
you install must be provided in the <strong>native Linux formats</strong> listed below. If a mod
author only ships <code>.dll</code> files, that mod <strong>will not work</strong> on this Linux build —
there is no <code>.dll</code>-to-<code>.so</code> compatibility layer, emulation, or shim of any kind.</p>
<markdown-accessiblity-table data-catalyst="">
Mod Type | Windows Format | Required Linux Format
-- | -- | --
Lua mods | scripts/main.lua | scripts/main.lua — identical, no changes needed (Lua is script-based and platform-independent)
C++ mods | dlls/main.dll | libs/main.so — must be a natively compiled Linux shared object (.so), not a .dll

</markdown-accessiblity-table>
<p dir="auto"><strong>Key points to remember:</strong></p>
<ul dir="auto">
<li>The mod folder for C++ mods on Linux is called <strong><code>libs/</code></strong>, not <code>dlls/</code> (this is different from Windows!).</li>
<li>A C++ mod must be <strong>recompiled from source specifically for Linux</strong> (e.g. with GCC or Clang, producing an ELF <code>.so</code> file). Simply renaming a <code>.dll</code> to <code>.so</code> will <strong>not</strong> work — the internal binary format is completely different.</li>
<li>Directory names inside a mod folder are <strong>case-sensitive</strong> on Linux (<code>scripts</code>, <code>libs</code> — always lowercase), unlike Windows where case does not matter.</li>
<li>If your mod only provides a <code>dlls/</code> folder with a <code>.dll</code> inside, UE4SS will <strong>not detect or load it at all</strong> on Linux — you will need the mod author (or yourself, if you have the source) to produce a Linux <code>.so</code> build.</li>
</ul>
[...]

Issue #885, opened 2026-07-15
Add server-side mod support for Palworld 1.0 on Linux


Alarisco:
## Is your feature request related to a problem? Please describe

Palworld 1.0 introduced official server-side mod support, but it currently appears to be limited to Windows dedicated servers.

I am using `thijsvanloef/palworld-server-docker:latest`, which runs the Linux dedicated server, so mods that require server-side installation cannot currently be used.

## Describe the solution you'd like

Are there any plans to add server-side mod support to this Docker image once Pocketpair enables it for Linux dedicated servers?

Ideally, the image could provide support for the official Palworld mod system, including:

- `Mods/Workshop`
- `PalModSettings.ini`
- Enabling and managing installed server-side mods
- Persistent mod files through Docker volumes
- Optional automatic Workshop mod installation or updates

## Describe alternatives you've considered

The current alternatives are:

- Using client-side-only mods
- Running a Windows dedicated server instead of the Linux Docker image
- Manually modifying the container, which would be unsupported and could break after updates

## Additional context

The main use case is installing server-side mods such as QualityOfLife, including features like expanding the base radius.

I understand that this may depend entirely on Pocketpair first releasing official server-side mod support for Linux. I mainly wanted to ask whether support is planned for this image once that becomes possible.

Thanks for maintaining the project.

## Feature Report Checklist

- [ ] I am willing to implement this feature

thijsvanloef:
Hi @Alarisco ! Thanks for creating the feature request, yes I plan on adding modding support, however it requires a bit more work then expected, since most of the mods require UE4SS which requires another translation layer like proton/wine. It is something I would like to include in the image, but have not had the time to tackle.

niklas2233:
Here is a repo for Palworld server with windows version through wine with mod support. If you run the server with -nosteam you can copy the world files to the new server without losing progress
https://github.com/ripps818/docker-palworld-dedicated-server-wine

MusclePr:
@niklas2233

Using the repository you recommended as a reference, I have implemented mod support utilizing Wine+WineTricks+UE4SS.

- https://github.com/MusclePr/palworld-server-docker/tree/feature/SupportsMods

@thijsvanloef 

I plan to submit a pull request, but since this involves a major feature addition—including Wine—I would like to organize the information a bit more before doing so.

I would like to align the design to make it easy to switch to Linux UE4SS. ( #906 )

MusclePr:
I am working on #914.

[...]

PR #913, opened 2026-07-29
feat: add UE4SS mod loader

lhw:
## Context 
Bundle the Linux UE4SS build for amd64 servers, arm64 is disabled for now (needs patched Box64)
Implementation for #906 and potentially  #885, though it does not use steam workshop mods yet and is native instead of wine/windows.

## Choices
Native should always be preferable, but tbh I also didn't check that there was already work being done.
For arm64/Box64 support I already have something local but there are some issues with the standard box64 binary, mostly that it doesn't have enough callback entries in it's table, ue4ss seems to request 32. But I will look into if there is an easier/better option to make it work.

## Test instructions

I did include a small test Lua mod to show that the loading works fine. General mod support depends on the mod itself though. I only tried two so far.

## AI Usage Disclosure

Please read the [AI Usage Guidelines for Contributors](AI_GUIDELINES.md) before submitting this PR.
If you used an AI tool to assist in writing or refactoring code for this PR, please disclose it below.

- [x] I used OpenAI Codex to assist in modifications for this PR.
- [x] I have manually reviewed the generated code and tested the container build locally.


lhw:
Ah damn for proper steam workshop support palschema needs to be compiled for linux. A little bit of extra work but definitely doable.

lhw:
I tried to get PalSchema to compile for ue4ss-linux and while I can find most of the symbols to hook into, some important ones are missing or would need to be identified during runtime. So I think it makes sense for now to focus on the wine/proton + windows server approach and I'll update if I can find a reasonably good approach to get all the symbols at any time I'll update the PR.

Note that the PR as is works fine for LUA mods, just not cpp ones

thijsvanloef:
Closing in favor of https://github.com/thijsvanloef/palworld-server-docker/pull/914

PR #917, opened 2026-08-01
Update documentation for community server public port

Dashboy1998:
## Context 

Issue in documentation noticed by @Cryator in regards to the community server that to use a different public port you must change `PUBLIC_PORT`

## Choices

Updated documentation to make it clear you need to change `PUBLIC_PORT` for a different port. 

Also commented out port forwarding for the restful API as that should not be listed as enabled. Truthfully this should probably be removed from the example and listed under an advance user option.

## Test instructions

N/A

## AI Usage Disclosure

Please read the [AI Usage Guidelines for Contributors](AI_GUIDELINES.md) before submitting this PR.
If you used an AI tool to assist in writing or refactoring code for this PR, please disclose it below.

- [ ] I used [Tool Name] to assist in modifications for this PR.
      I have manually reviewed the generated code and tested the container build locally.


gammamonbiscuit:
This makes we wonder that, while having `PUBLIC_PORT` added, should the example also have `PUBLIC_IP` and explain its usage?

Although this sounds like an advanced option, it is necessary for those behind CGNAT, need to put their server behind a reverse proxy, and want it available in community listing.

Dashboy1998:
> This makes we wonder that, while having `PUBLIC_PORT` added, should the example also have `PUBLIC_IP` and explain its usage?
>
> Although this sounds like an advanced option, it is necessary for those behind CGNAT, need to put their server behind a reverse proxy, and want it available in community listing.

I agree that's probably a good idea. I'm not sure how to best phrase when to use PUBLIC_IP as it's generally not need and I've never tested it behind a reverse proxy or CGNAT.

